PR #8, opened 2020-02-01
Implement a generic verification protocol for the string type claims.

grosch:


Sources/JWTKit/JWTClaims.swift
```diff
@@ -1,33 +1,41 @@
...
-    
+
     /// See `JWTClaim`.
     public init(value: String) {
         self.value = value
     }
+
+    /// Throws an error if the claim's value is not one of the accepted values.
+    /// - Parameter oneOf: The accepted values.
+    public func verify(claim name: String, oneOf: String...) throws {
```
gwynne: This method can be simplified to:

```swift
guard oneOf.contains(name) else {
    throw JWTError.claimVerificationFailure(name: name, reason: "Not of of the allowed values.")
}
```

grosch:
Good catch!  I just pushed an update to switch to that.

Sources/JWTKit/JWTClaims.swift
grosch: Fixed

siemensikkema:
I generally approve of reducing code duplication where possible! But all of the string claims are now interchangeable instead of separate types (typealias is just syntax sugar). 
The verify methods you added (which in itself goes beyond just reducing duplication in JWTKit) produce an error that does not indicate which claim they are from. That might be unhelpful during debugging and especially when communicating the errors to the end user.
So I'd like to see what @tanner0101 thinks before approving ...

gwynne:
@siemensikkema Excellent point! In this case, that could be solved by making `StringClaim` a protocol, adding the `verify` implementations in an `extension`, and having the four concrete claims conform to the protocol (the initializer can and should be required by the protocol, but sadly an implementation for it can't be provided in an extension). It doesn't de-duplicate as _much_ code this way, but it still does a decent amount.

grosch:
Is there really any value to these being separate types though?  There's absolutely no difference in the implementation.  It's just a name.

@gwynne Maybe I'm missing something, but i'm not seeing how making a StringClaim protocol reduces the code at all.

grosch:
OK, so I put it back to being specific types, even though I'm not seeing the real value to that, and instead have them implement an `JWTClaimVerifiable` protocol to include the three verification methods.  Not as clean feeling to me, but hopefully it's an acceptable change to everyone now.

Sources/JWTKit/JWTClaimVerifiable.swift
```diff
@@ -0,0 +1,27 @@
...
+
+    func verify(oneOf desired: Value...) throws
+    func verify(oneOf desired: [Value]) throws
+    func verify(is desired: Value) throws
+}
+
+public extension JWTClaimVerifiable {
+    func verify(is desired: Value) throws {
+        guard desired == self.value else {
+            throw JWTError.claimVerificationFailure(name: Self.claimName, reason: "Issuer is incorrect.")
```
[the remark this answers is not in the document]
grosch: Shoot, I thought I had included that one with the last push.  It's fixed now.  Thanks for catching!

siemensikkema:
A way to further reduce code duplication would be to use generics, like so:
```swift
protocol StringClaimTag {
    static var name: String { get }
}

typealias AudienceClaim = StringClaim<StringClaim.AudienceTag>
typealias IssuerClaim = StringClaim<StringClaim.IssuerTag>
typealias SubjectClaim = StringClaim<StringClaim.SubjectTag>

struct StringClaim<Claim: StringClaimTag> {
    struct AudienceTag: StringClaimTag {
        static var name: String { "aud" }
    }

    struct IssuerTag: StringClaimTag {
        static var name: String { "iss" }
    }

    struct SubjectTag: StringClaimTag {
        static var name: String { "sub" }
    }

    let value: String
    func verify(equals: String) {
        guard value == equals else {
            throw JWTError.claimVerificationFailure(name: Claim.name, reason: "Value is incorrect.")
        }
    }
}
```

grosch:
The nesting wasn't working so the tags aren't under the StringClaim struct.

siemensikkema:
@tanner0101 do you have an opinion on which way to go with this? Options that we've explored are:
- duplication
  - pros: no public namespace solution
  - cons: duplication
- a single StringClaim with type-aliases
  - pros: no duplication
  - cons: types are interchangeable
- using a protocol
  - pros: not duplicating verify functionality
  - cons: duplicating types
- using a generic StringClaim with "tag" types.
  - pros: no duplication, similar approach could be used for date backed claims
  - cons: public namespace polution with tags

gwynne:
The reason to have truly separate types is to leverage the type system for its intended purpose - to prevent providing an instance of X when an instance of type Y is required. With distinct types, this can be caught by the compiler during development rather than becoming a potentially hidden-until-production runtime error.

grosch:
I understand type checking. The point was as implemented it didn’t do that. The rules were 100% duplicates of each other and nothing would prevent somebody from defining their aid type as a SubjectClaim. It would be like saying swift should provide FirstNameString and LastNameSteing types because that’s the expected usage.

gwynne:
```swift
struct FirstNameString { var value: String }
struct LastNameString { var value: String }
func useFirstName(_ name: FirstNameString) { print("Name: \(name.value)") }
useFirstName(LastNameString(value: "Lastname"))
// error: cannot convert value of type 'LastNameString' to expected
// argument type 'FirstNameString'
```

While it's true that it's still fairly easy to mix and match the types as they were written before, the compiler can still provide at least some type safety. That goes away entirely with typealiases:

```swift
struct NameString { var value: String }
typealias FirstNameString = NameString
typealias LastNameString = NameString
func useFirstName(_ name: FirstNameString) { print("Name is \(name.value)") }
useFirstName(LastNameString(value: "Lastname")) // prints "Name is Lastname"
```

(And yes, the `ExpressibleByStringLiteral` conformances further muddy the waters, but there's still at least some safety provided.)

grosch:
@tanner0101 Can you comment or merge please?

grosch:
Too contentious.  I'll just put the verify stuff into my projects.

PR #31, opened 2020-07-16
Boring SSL update

0xTim:
Brings in the latest version of BoringSSL (#31).


0xTim:
@tanner0101 this test is failing on master too - looks like there might actually be a bug. From the brief look at it, it's looks like the payload is correct but the header and signature are different

tanner0101:
Looks like the JWT header key sorting changed. So a bug in the test technically. 

Screenshot from 2020-07-16 12-12-19

I set the header encoder to sort keys but I'm not sure that's the right solution. The keys don't really need to be sorted and that requires CPU. 

Really the deterministic encoding just makes testing easier, so this should probably go behind some sort of flag.

0xTim:
Yeah I agree having the header keys sorted in prod code is not the way forward. My preference would be to refactor the tests so that they don't care about JSON ordering - since this is what happens in real life. As a first pass to get this and #32 unblocked we could check both variants in the test. Since there are only two keys there are only two possible values so it's not unreasonable

tanner0101:
These changes are now available in [4.0.0-rc.1.6](https://github.com/vapor/jwt-kit/releases/tag/4.0.0-rc.1.6)

PR #52, opened 2021-02-17
Make CJWTKitBoringSSL implementation only

### LINKED ISSUE #51: Use @_implementationOnly to hide BoringSSL
### Describe the bug

When building JWTKit alongside any other project using a vendored copy of BoringSSL or OpenSSL, there is a risk of getting C++ ODR violations: see kylebrowning/APNSwift#102 for an example.

### To Reproduce

1. Open an Xcode that contains a snapshot version of Swift 5.4.
2. Depend on APNSwift.
3. Try to build.

### Expected behavior

Builds to succeed

### Additional context

The ODR violations are caused by the fact that module imports in Swift are assumed to be part of the importing module's API. That is, if module A imports module B, unless any action is taken all other modules that depend on A implicitly need to see the module definition of B.

For C/C++ modules, this has the effect of passing the header files include path of CJWTKitBoringSSL to all downstream builders. If those downstream builders attempt to import CNIOBoringSSL as well (or any other BoringSSL/OpenSSL vendored project not using `@_implementationOnly`) then the build will fail.

This is exacerbated by SwiftNIO having been cautious about its use of this flag, though we plan to ship a release shortly that contains our fix for this issue (apple/swift-nio-ssl#266). The fact that NIO was using `@_implementationOnly` has hidden this issue with JWTKit for the moment, but JWTKit should also be using it.

0xTim:
Makes CJWTKitBoringSSL implementation only to avoid leaking C types from BoringSSL. Resolves #51 

Also updates BoringSSL to f6bd54efbcafcf4625ce99b5f702dc4850b0ca50


siemensikkema (approved):
LGTM, thanks for digging into this one!

tanner0101:
These changes are now available in [4.2.1](https://github.com/vapor/jwt-kit/releases/tag/4.2.1)

PR #59, opened 2021-07-16
Add `is_private_email` claim to `AppleIdentityToken`

gargs:
This adds the missing `is_private_email ` claim to `AppleIdentityToken`. This is helpful to determine whether the user is using a proxy email address.


0xTim:
@gargs  thanks! Could you add a link to the docs where it's described?

gargs:
Sorry, here it is - 

https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple

The documentation states that the value is either 'true' or 'false', but in my testing (on sandbox), the value was missing if false, and true otherwise.

0xTim (approved):

PR #61, opened 2021-09-19
Fix typo for `JWTError.signatureVerificationFailed` case

davdroman:
Just something I noticed while using the library. Unless there's another way to do, this would be a breaking change, so if semver is to be strictly followed it'd involve a major version bump. Up to you :)

### DIFF
--- a/Sources/JWTKit/JWTError.swift
+++ b/Sources/JWTKit/JWTError.swift
@@ -4,7 +4,7 @@ public enum JWTError: Error, CustomStringConvertible, LocalizedError {
     case claimVerificationFailure(name: String, reason: String)
     case signingAlgorithmFailure(Error)
     case malformedToken
-    case signatureVerifictionFailed
+    case signatureVerificationFailed
     case missingKIDHeader
     case unknownKID(JWKIdentifier)
     case invalidJWK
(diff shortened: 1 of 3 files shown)


0xTim:
Oof yeah that's not something we can add it without a breaking change unfortunately. We'll need to save this for a major version bump

0xTim:
@ptoffy want to merge this into your branch?

ptoffy:
> @ptoffy want to merge this into your branch?

This is fixed in my branch already 😅

Sources/JWTKit/JWTError.swift
```diff
@@ -4,7 +4,7 @@ public enum JWTError: Error, CustomStringConvertible, LocalizedError {
     case claimVerificationFailure(name: String, reason: String)
     case signingAlgorithmFailure(Error)
     case malformedToken
-    case signatureVerifictionFailed
+    case signatureVerificationFailed
```
gwynne, line 7: This doesn't have to be a semver-major change. Just add this:
```swift
extension JWTError {
    @available(*, deprecated, renamed: "signatureVerificationFailed")
    public var signatureVerifictionFailed; JWTError { .signatureVerificationFailed }
}
```

0xTim: But @davdroman if you want to update it we can add it to v4

davdroman:
Succeeded by #103.

PR #63, opened 2021-09-28
Added tvOS support

mbutan:
Adds support for tvOS


0xTim:
I have no objections to this 👍

0xTim:
Original comment:

> It would be great to have support for a tvOS platform. Already tested solution and seems to work stable.

Best.

0xTim (approved):

PR #66, opened 2022-02-26
Expose the curve on ECDSA Keys

danramteke:
Exposes the `Curve` on `ECDSAKey`s

### DIFF
--- a/Sources/JWTKit/ECDSA/ECDSAKey.swift
+++ b/Sources/JWTKit/ECDSA/ECDSAKey.swift
@@ -16,6 +16,19 @@ public final class ECDSAKey: OpenSSLKey {
                 return NID_secp521r1
             }
         }
+      
+        public init?(cName: Int32) {
+            switch cName {
+            case NID_X9_62_prime256v1:
+                self = .p256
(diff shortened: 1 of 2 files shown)


0xTim (approved):
Thanks!

0xTim:
Original comment:

This PR parses the `Curve` from an `ECDSAKey`.

## Implementation

1. Add new init on `ECDSAKey.Curve` to initialize from a `cName` (an `Int32`), just as a `Curve` can be converted to a `cName`. 
2. Use BoringSSL to fetch the curve name from the `group`. Init the `Curve` from the returned `Int32` using the new initializer. This method returns `nil` on failure, just like the sibling `var parameters: Paramaters?.

## Motivation

I wish to store only PEMs in our secrets vault, for simplicity. Extracting the Curve is one step towards this flow:

```
let myKey = ECDSAKey.private(string: Environment.get("MY_KEY_PEM")!)
let myKeyPublic = ECDSAKey.public(string: Environment.get("MY_KEY_PUBLIC_PEM")!)

// ... 

JWKS(keys: JWK(ecdsa: myKeyPublic))
```

PR #73, opened 2022-06-09
FIX Issue #46

MFranceschi6:
Hi from [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517)

an jwk representing an ecdsa key is represented with the parameters x and y in base64url-encoded form.

The issue was caused by the fact that the parameters were converted using `Data(base64Encoded: String)` which doesn't decode an base64url string

### DIFF
--- a/Sources/JWTKit/Utilities/BigNumber.swift
+++ b/Sources/JWTKit/Utilities/BigNumber.swift
@@ -16,8 +16,8 @@ class BigNumber {
         CJWTKitBoringSSL_BN_free(self.c);
     }
 
-    public static func convert(_ bnBase64: String) -> BigNumber? {
-        guard let data = Data(base64Encoded: bnBase64) else {
+    public static func convert(_ bnBase64Url: String) -> BigNumber? {
+        guard let data = bnBase64Url.data(using: .utf8)?.base64URLDecodedBytes() else {
             return nil
         }
 
--- a/Tests/JWTKitTests/JWTKitTests.swift
+++ b/Tests/JWTKitTests/JWTKitTests.swift
@@ -290,6 +290,44 @@ class JWTKitTests: XCTestCase {
         let foo = try signers.verify(jwt, as: Foo.self)
         XCTAssertEqual(foo.bar, 42)
     }
+
+    func testVerifyingECDSAKeyUsingJWKBase64URL() throws {
+        struct Foo: JWTPayload {
+            var bar: Int
+            func verify(using signer: JWTSigner) throws { }
+        }
+                
+        // ecdsa key
+        let x = "BRSikX2rNlAhJD5mw1KCEAIKK3_yttL_zvo9IXvD59o"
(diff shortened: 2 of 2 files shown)


MFranceschi6:
Fix issue #46

MFranceschi6:
If someone could request a review

Sources/JWTKit/Utilities/BigNumber.swift
```diff
@@ -27,12 +27,12 @@ class BigNumber {
         return BigNumber(c)
     }
 
-    public func toBase64(_ size: Int = 1000) -> String {
+    public func toBase64URL(_ size: Int = 1000) -> String {
```
0xTim, line 30: We'll need to keep this around since it's a public function (to avoid breaking the public API). Given there are other instances when a Base 64 string might be useful I suggest we just create a new function with the correct behaviour

Sources/JWTKit/Utilities/BigNumber.swift
```diff
@@ -16,8 +16,8 @@ class BigNumber {
         CJWTKitBoringSSL_BN_free(self.c);
     }
 
-    public static func convert(_ bnBase64: String) -> BigNumber? {
-        guard let data = Data(base64Encoded: bnBase64) else {
+    public static func convert(_ bnBase64Url: String) -> BigNumber? {
```
0xTim, line 19: As below - this is technically non-breaking but I'd prefer to create a new function that is explicitly for Base64URL strings to avoid breaking stuff people are using this for and switching our ECDSA implementation over to that

0xTim (changes requested):
Thanks for spotting this (and adding a test!)

We'll need to make some changes to avoid making this a breaking change

Sources/JWTKit/Utilities/BigNumber.swift
MFranceschi6: Can I propose
```swift
public convenience init?(base64URL: String)
```
for the `BigNumber` class?

0xTim: Yep that sounds good

Sources/JWTKit/Utilities/BigNumber.swift
```diff
@@ -16,6 +16,8 @@ class BigNumber {
         CJWTKitBoringSSL_BN_free(self.c);
     }
 
+
+
```
0xTim, line 20: Nit - extra space

0xTim (changes requested):
One small nit then good to go

0xTim (approved):
Thanks!

0xTim:
Original comment:
Hi from [RFC7517](https://datatracker.ietf.org/doc/html/rfc7517)

an jwk representing an ecdsa key is represented with the parameters x and y in base64url-encoded form.

The issue was caused by the fact that the parameters were converted using `Data(base64Encoded: String)` which doesn't decode an base64url string

PR #75, opened 2022-06-24
Add JWS Support

0xTim:
@Lukasa here's a horrible written spike for verifying an X5c chain in a JWS - could you take a quick look when you get a chance and make sure I'm not doing anything stupid with the C code? (I know I need to make the cert types generic based off the `alg` field

### DIFF
--- a/Sources/JWTKit/JWTSigners.swift
+++ b/Sources/JWTKit/JWTSigners.swift
@@ -144,7 +144,7 @@ public final class JWTSigners {
         // Verify the chain
         // The first cert is used to sign the JWS
         // Each subsequent cert should be used to certify the previous one
-        // For the last cert we can find the signer for the KID
+        // For the last cert we use the root cert passed in
         // https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6
         for (index, certificate) in x5c.enumerated() {
             if index == 0 {
--- a/Sources/JWTKit/JWTSigners.swift
+++ b/Sources/JWTKit/JWTSigners.swift
@@ -110,14 +110,14 @@ public final class JWTSigners {
         try JWTParser(token: token).payload(as: Payload.self)
     }
     
-    public func verifyJWS<Payload>(
+    public func verifyJWSWithX5C<Payload>(
         _ token: String,
         as payload: Payload.Type = Payload.self,
         rootCert: String
     ) throws -> Payload
         where Payload: JWTPayload
     {
-        try self.verifyJWS([UInt8](token.utf8), as: Payload.self, rootCert: [UInt8](rootCert.utf8))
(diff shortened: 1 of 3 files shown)


betaphi:
@0xTim @Lukasa I think this draft is very useful for many users. Are there any plans to release it in the near future?

Sources/JWTKit/JWTSigners.swift
```diff
@@ -108,8 +109,99 @@ public final class JWTSigners {
...
+        // The first cert is used to sign the JWS
+        // Each subsequent cert should be used to certify the previous one
+        // For the last cert we use the root cert passed in
+        // https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6
+        for (index, certificate) in x5c.enumerated() {
+            if index == 0 {
+                continue
+            }
+            
+            if index == x5c.count {
```
WFT, line 154:
```suggestion
            if index == x5c.count - 1 {
```
Otherwise this code won't be run, since `x5c.enumerated().last?.offset == x5c.count - 1`. This means that the root cert is never checked. So any valid chain is verified, even if it's not using the trusted root cert.

WFT (commented):
Hey @0xTim, I’m the guy from the forum post that kicked this off. I’ve been tooling around with this for a little while before going to production and the integration is great! I noticed what I believe to be a pretty serious off-by-one error which I’ve highlighted below.

I do have also a few questions: (1) is there a path towards getting this merged into main? (2) if so, is there anything I can do to help? I’m not very familiar with BoringSSL or I’d give a review of the C code too. I’ve written some tests, which I’m happy to share if they’d be helpful. I’d be happy to do any other work I can to help.

Sources/JWTKit/JWTSigners.swift
```diff
@@ -108,8 +109,99 @@ public final class JWTSigners {
...
+                defer { CJWTKitBoringSSL_X509_free(x5092) }
+                
+                let pkey = CJWTKitBoringSSL_X509_get_pubkey(x509)
+                defer { CJWTKitBoringSSL_EVP_PKEY_free(pkey) }
+                
+                guard CJWTKitBoringSSL_X509_verify(x5092, pkey) == 1 else {
+                    throw JWTError.generic(identifier: "JWS", reason: "Certificate verification failed")
+                }
+            }
+        }
```
Lukasa, line 193: To do cert verification it's better not to reinvent `X509_STORE_CTX`. Prefer using `X509_verify_cert`, which will force you to construct the `X509_STORE_CTX` appropriately for validation.

WFT:
Hey, I’ve taken a pass at this using the X509_STORE/X509_verify_cert API. That’s over here: #84

0xTim:
Closing as this was merged in #84

PR #76, opened 2022-07-07
Add experimental watchOS support

thebabufrik:
Adds experimental watchOS support by providing a watchOS platform to match the minimum version of the dependencies so it will compile


0xTim (approved):
Thanks!

PR #77, opened 2022-07-14
Support fetch subjects from RSAKey and create JWK From Pem.

AFutureD:
This PR support 2 things:
1. Now we can get subjects from a RSA key, such as modulus and publicExponent.
2. JWK can create from Pem file more easily.

**Examples**:

1. Get subjects form RSA key

```Swift

let key = try RSAKey.public(pem: "")
guard let modulus = key.modulus else {
    return
}
        
let base64Encoded = modulus.base64EncodedString()
let base64URLEncoded = modulus.base64URLEncodedString()

```

2. Create JWK from pem

```Swift

let key = RSAKey(modulus: rsaModulus, exponent: "AQAB", privateExponent: rsaPrivateExponent)

let jwk = JWK.rsa(key, .rs256, identifier: "1234")

let jsonEncoder = JSONEncoder()
jsonEncoder.dateEncodingStrategy = .secondsSince1970

let jsonData = try jsonEncoder.encode(jwk)
let json = String(data: jsonData, encoding: .utf8)

```


0xTim:
(This is on my TODO list to review - but heads up we're planning on moving away from vending our own copy of BoringSSL in favour of sitting on top of Swift Crypto directly. So I'm hesitant to add new APIs that call straight through to the C code that we can't easily replace with Swift Crypto)

AFutureD:
Great! Using Swift Crypto is a good choice, I'll check if I can use it instead.

0xTim:
@ptoffy is this available in the new version?

ptoffy:
> @ptoffy is this available in the new version?

Unfortunately not, but we can see to its addition. Swift Crypto doesn't currently provide that API but we can probably work around that

AFutureD:
This feature may supported in version 5?
I found this commit in branch jwtkit-5. https://github.com/vapor/jwt-kit/commit/7ceeaa7d631697bc1354474edb762c50b621d801#diff-ce6a5cb263b10684577a112085166f0c803df548d6430cbe4cec06f6d979f6a0

AFutureD:
close as implemented in version 5.
https://github.com/vapor/jwt-kit/blob/main/Sources/JWTKit/RSA/RSAPrivateKey%2BASN1.swift

PR #78, opened 2022-08-01
Allow empty signature

baarde:
When encoding an Unsecured JWT with the "none" algorithm, the encoded signature is the empty string.


0xTim:
Just to confirm - if we merge the other PR we still need this right?

baarde:
No, sorry for the confusion. The commit in this PR is included in #79. If you're OK with that PR, we can close this one.

PR #79, opened 2022-08-01
Add `JWTSigner.unsecuredNone`

baarde:
This PR adds support for the "none" algorithm described in [Section 6 of RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-6).

### DIFF
--- a/Sources/JWTKit/JWTParser.swift
+++ b/Sources/JWTKit/JWTParser.swift
@@ -8,7 +8,8 @@ struct JWTParser {
     init<Token>(token: Token) throws
         where Token: DataProtocol
     {
-        let tokenParts = token.copyBytes().split(separator: .period)
+        let tokenParts = token.copyBytes()
+            .split(separator: .period, omittingEmptySubsequences: false)
         guard tokenParts.count == 3 else {
             throw JWTError.malformedToken
         }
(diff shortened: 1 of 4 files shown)


baarde:
The name `.unsecuredNone` is verbose on purpose, as most users probably don't want to use this.

Alternatives:
* `.none` could cause confusion between `Optional<JWTSigner>.some(.none)` and  `Optional<JWTSigner>.none`.
* `.unsecured` may be hard to discover using autocomplete.
* `.noneUnsecured` is fine by me.
* Rejecting the PR and letting the user implement this is probably also fine (tough [Section 8 of RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519#section-8) states that an implementation MUST support "none"). See #78.

0xTim (changes requested):
Thanks for adding this! I agree we should definitely support it and I think the name is good. 

Could you expand the test to ensure the string version of the token matches? Since it's all static should be fine as it won't break. Just in case we break it in the future and start trying to sign it

baarde:
Comparing the encoded token is not possible as the order of the keys is non-deterministic (unless we add the `.sortedKeys` option to `JSONEncoder`).
```swift
let data = "eyJhbGciOiJub25lIiwidHlwI..."
let token = try signer.sign(payload)
XCTAssertEqual(token, data) // ❌
```

However, we can check that the token's signature is empty:

```swift
XCTAssertTrue(token.hasSuffix("."))
```

0xTim:
Ahh you're right. What we can do though is verify the token when the token string is not signed

0xTim (approved):
Thanks!

PR #82, opened 2022-08-18
Add support for Octet Key Pairs with the EdDSA algorithm using the ED25519 curve

### LINKED ISSUE #50: EdDSA Support
Please consider supporting Ed25519 signature.

mgratzer:
This implements support for Octet Key Pairs ([RFC8037](https://www.rfc-editor.org/rfc/rfc8037.html#section-2)) with the EdDSA algorithm and ED25519 curves. It closes #50.

### DIFF
--- a/Sources/JWTKit/ECDSA/ECDSAKey.swift
+++ b/Sources/JWTKit/ECDSA/ECDSAKey.swift
@@ -5,6 +5,7 @@ public final class ECDSAKey: OpenSSLKey {
         case p256 = "P-256"
         case p384 = "P-384"
         case p521 = "P-521"
+		case ed25519 = "Ed25519"
 
         var cName: Int32 {
             switch self {
@@ -14,6 +15,8 @@ public final class ECDSAKey: OpenSSLKey {
                 return NID_secp384r1
             case .p521:
                 return NID_secp521r1
+			case .ed25519:
+				return NID_ED25519
             }
         }
       
@@ -25,6 +28,8 @@ public final class ECDSAKey: OpenSSLKey {
                 self = .p384
             case NID_secp521r1:
                 self = .p521
+			case NID_ED25519:
+				self = .ed25519
             default:
                 return nil
             }
--- a/Sources/JWTKit/ECDSA/ECDSAKey.swift
+++ b/Sources/JWTKit/ECDSA/ECDSAKey.swift
@@ -2,10 +2,15 @@
 
 public final class ECDSAKey: OpenSSLKey {
     
-    @available(*, deprecated, renamed: "JWK.Curve")
-    public typealias Curve = JWK.Curve
+    public enum Curve: String, Codable {
+        case p256 = "P-256"
+        case p384 = "P-384"
+        case p521 = "P-521"
+        case ed25519 = "Ed25519"
+        case ed448 = "Ed448"
+    }
     
-    public static func generate(curve: JWK.Curve = .p521) throws -> ECDSAKey {
+    public static func generate(curve: Curve = .p521) throws -> ECDSAKey {
         guard let c = CJWTKitBoringSSL_EC_KEY_new_by_curve_name(curve.cName) else {
             throw JWTError.signingAlgorithmFailure(ECDSAError.newKeyByCurveFailure)
         }
@@ -47,7 +52,7 @@ public final class ECDSAKey: OpenSSLKey {
         self.c = c
     }
     
-    public convenience init(parameters: Parameters, curve: JWK.Curve = .p521, privateKey: String? = nil) throws {
+    public convenience init(parameters: Parameters, curve: Curve = .p521, privateKey: String? = nil) throws {
         guard let c = CJWTKitBoringSSL_EC_KEY_new_by_curve_name(curve.cName) else {
             throw JWTError.signingAlgorithmFailure(ECDSAError.newKeyByCurveFailure)
         }
@@ -79,10 +84,10 @@ public final class ECDSAKey: OpenSSLKey {
         CJWTKitBoringSSL_EC_KEY_free(self.c)
     }
   
-    public var curve: JWK.Curve? {
+    public var curve: Curve? {
         let group: OpaquePointer = CJWTKitBoringSSL_EC_KEY_get0_group(self.c)
         let cName = CJWTKitBoringSSL_EC_GROUP_get_curve_name(group)
-        return JWK.Curve(cName: cName)
+        return Curve(cName: cName)
     }
     
     public var parameters: Parameters? {
@@ -104,7 +109,7 @@ public final class ECDSAKey: OpenSSLKey {
     }
 }
 
-extension JWK.Curve {
+extension ECDSAKey.Curve {
     var cName: Int32 {
         switch self {
         case .p256:
--- a/Sources/JWTKit/ECDSA/ECDSAKey.swift
+++ b/Sources/JWTKit/ECDSA/ECDSAKey.swift
@@ -1,6 +1,9 @@
 @_implementationOnly import CJWTKitBoringSSL
 
 public final class ECDSAKey: OpenSSLKey {
+	
+	@available(*, deprecated, renamed: "JWK.Curve")
+	public typealias Curve = JWK.Curve
     
 	public static func generate(curve: JWK.Curve = .p521) throws -> ECDSAKey {
         guard let c = CJWTKitBoringSSL_EC_KEY_new_by_curve_name(curve.cName) else {
--- a/Sources/JWTKit/ECDSA/ECDSAKey.swift
+++ b/Sources/JWTKit/ECDSA/ECDSAKey.swift
@@ -5,7 +5,7 @@ public final class ECDSAKey: OpenSSLKey {
 	@available(*, deprecated, renamed: "JWK.Curve")
 	public typealias Curve = JWK.Curve
     
-	public static func generate(curve: JWK.Curve = .p521) throws -> ECDSAKey {
+    public static func generate(curve: JWK.Curve = .p521) throws -> ECDSAKey {
         guard let c = CJWTKitBoringSSL_EC_KEY_new_by_curve_name(curve.cName) else {
             throw JWTError.signingAlgorithmFailure(ECDSAError.newKeyByCurveFailure)
         }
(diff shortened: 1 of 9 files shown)


Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -1,37 +1,8 @@
 @_implementationOnly import CJWTKitBoringSSL
 
 public final class ECDSAKey: OpenSSLKey {
-    public enum Curve: String, Codable {
```
tonyarnold, line 4: Can this be replaced with a typealias to ``JWK.Curve`` that is marked as deprecated/renamed to ensure that existing code doesn't break?

0xTim:
Original comment:
This PR implements support for Octet Key Pairs ([RFC8037](https://www.rfc-editor.org/rfc/rfc8037.html#section-2)) with the EdDSA algorithm and ED25519 curves. It closes #50.

I aligned the implementation with the existing code for `ECDSA` as close as possible. 

I used `swift-crypto`, which limits us to the `ED25519` curve (which seems to be the more standard curve for EdDSA?), and could not add support for the `ED448` curve (which would be supported by `BoringSSL`).

⚠️ Potential breaking change: 
~~I renamed `ECDSAKey.Curve` to `JWK.Curve` because `Curve` is no longer limited to `ECDSA`.~~ https://github.com/vapor/jwt-kit/pull/82#pullrequestreview-1080300524

Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -1,37 +1,11 @@
...
-                return nil
-            }
-        }
-    }
+	
+	@available(*, deprecated, renamed: "JWK.Curve")
+	public typealias Curve = JWK.Curve
     
-    public static func generate(curve: Curve = .p521) throws -> ECDSAKey {
+	public static func generate(curve: JWK.Curve = .p521) throws -> ECDSAKey {
```
0xTim, line 8: There looks to be some weird indentation here

Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -73,7 +47,7 @@ public final class ECDSAKey: OpenSSLKey {
         self.c = c
     }
     
-    public convenience init(parameters: Parameters, curve: Curve = .p521, privateKey: String? = nil) throws {
+	public convenience init(parameters: Parameters, curve: JWK.Curve = .p521, privateKey: String? = nil) throws {
```
0xTim, line 50: As above

Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -129,3 +103,37 @@ public final class ECDSAKey: OpenSSLKey {
...
+			return NID_ED25519
+		case .ed448:
+			return NID_ED448
+		}
+	}
+  
+	init?(cName: Int32) {
+		switch cName {
+		case NID_X9_62_prime256v1:
+			self = .p256
```
0xTim, line 126: Indentation as above

0xTim (changes requested):
This looks great thanks! Sorry it's taken so long to get to!

There's just the issue of indentation, which needs to match the rest of the codebase and then we're good to go 🚀

Sources/JWTKit/ECDSA/ECDSAKey.swift
mgratzer: Seems my Xcode setup uses tabs 🤷🏼. Replaced all in https://github.com/vapor/jwt-kit/pull/82/commits/3c61d210de5a3be3a01965a4dd0e59923203d573 and https://github.com/vapor/jwt-kit/pull/82/commits/28a301a759f0b1a4b094b2c84c8866ce8924c569

0xTim (approved):
LGTM, thanks!

Sources/JWTKit/EdDSA/EdDSAKey.swift
```diff
@@ -0,0 +1,51 @@
...
+    
+    public init(x: String?, d: String? = nil, curve: JWK.Curve = .ed25519) throws {
+        try self.init(
+            publicKey: x.flatMap { $0.data(using: .utf8) }.map { Data($0.base64URLDecodedBytes()) },
+            privateKey: d.flatMap { $0.data(using: .utf8) }.map { Data($0.base64URLDecodedBytes()) },
+            curve: curve
+        )
+    }
+    
+    public init(publicKey: Data? = nil, privateKey: Data? = nil, curve: JWK.Curve = .ed25519) throws {
```
Lukasa, line 23: I'm a bit suspicious as to why we're letting users tell us about the curve. I suppose it's to let us do `ed448` in the future, but the fact that the API invites you to pass `.p256` to it is a bit weird.

Sources/JWTKit/EdDSA/EdDSAKey.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import Crypto
+
+public struct EdDSAKey {
```
Lukasa, line 4: It's a bit weird that this can be public _or_ private depending on how it's constructed.

Sources/JWTKit/EdDSA/EdDSAKey.swift
```diff
@@ -0,0 +1,51 @@
...
+    let keyPair: OctetKeyPair
+    var publicKey: Data? {
+        keyPair.publicKey
+    }
+    var privateKey: Data? {
+        keyPair.privateKey
+    }
+    let curve: JWK.Curve
+    
+    public init(x: String?, d: String? = nil, curve: JWK.Curve = .ed25519) throws {
```
Lukasa, line 15: You can technically construct this object with a public key and a private key that don't match. That seems bad.

Sources/JWTKit/JWTSigners.swift
```diff
@@ -229,6 +229,21 @@ private struct JWKSigner {
...
             }
+
+        case .octetKeyPair:
+            guard let algorithm = algorithm ?? self.jwk.algorithm else {
+                return nil
+            }
+                
+            switch algorithm {
+                case .eddsa:
+                    guard let key = try? EdDSAKey(x: self.jwk.x, d: self.jwk.privateExponent, curve: self.jwk.curve ?? .ed25519) else {
```
Lukasa, line 240: Is defaulting this to ed25519 right?

Sources/JWTKit/EdDSA/EdDSAKey.swift
mgratzer: I agree. That's weird on the level of specific keys. I scoped the supported curves for the particular key and added a mapping from the generic JWK definition, which could always be mismatched when parsed from JSON. c34e92a69dad2d4fcffcf51023a4e44f360cb179

Sources/JWTKit/JWTSigners.swift
mgratzer: Yes, it's the only curve supported for `EdDSA` right now. I made it more explicit with https://github.com/vapor/jwt-kit/pull/82#discussion_r993942834

Sources/JWTKit/EdDSA/EdDSAKey.swift
mgratzer: From https://www.rfc-editor.org/rfc/rfc8037.html#appendix-A I understood, that this can be the case depending on the use case (signing and verifying or verifying). In our specific use case, we provide a list of keys via an endpoint that does not contain the private key.

Do you think its beneficial to have two different vital types for more clarity?

Sources/JWTKit/EdDSA/EdDSAKey.swift
mgratzer: Is there a way to guarantee/verify that the keypair matches when provided like in https://www.rfc-editor.org/rfc/rfc8037.html#appendix-A

Sources/JWTKit/EdDSA/EdDSAKey.swift
Lukasa: I tend to think so, yes: it prevents you from attempting to do something with the wrong type of key.

Sources/JWTKit/EdDSA/EdDSAKey.swift
Lukasa: Yes: in the format you've linked to, the public key is provided as part of the private key.

Sources/JWTKit/EdDSA/EdDSAKey.swift
mgratzer: I changed the implementation. It's explicitly now if you create a private or public key. https://github.com/vapor/jwt-kit/pull/82/commits/12f9a890c5a5829b10c1ff262c128288e745e5cb

Sources/JWTKit/EdDSA/EdDSAKey.swift
mgratzer: Private and public keys are distinct now. The curve is not falling back to `.ed25519` anymore.
https://github.com/vapor/jwt-kit/pull/82/commits/12f9a890c5a5829b10c1ff262c128288e745e5cb

Sources/JWTKit/JWTSigners.swift
mgratzer: I removed the default fallback. Any unknown curve will not create a signer anymore. 
https://github.com/vapor/jwt-kit/pull/82/commits/12f9a890c5a5829b10c1ff262c128288e745e5cb

klemenkosir:
Any update on this? Why hasn't this been merged yet?

0xTim (approved):
Ok this looks good to me. Swift makes this release a bit difficult with new enum cases but I'm going to release it as a minor release

m-barthelemy:
I'm struggling to understand how to use this.
Supposing I'm doing the following:
```swift
    app.jwt.signers.use(.eddsa(
        try .private(
            x: "public stuff",
            d: "private stuff",
            curve: .ed25519
        )
    ))
```

How can I generate a key and then extract `x` and `d`?
I tried grabbing the values (without the headers) from the following commands:
```
openssl genpkey -algorithm ed25519 -out private.pem
openssl pkey -in private.pem -pubout
```

But my app throws `incorrectKeySize`.
The values generated by openssl are longer than the examples hardcoded in https://github.com/mgratzer/jwt-kit/blob/12f9a890c5a5829b10c1ff262c128288e745e5cb/Tests/JWTKitTests/JWTKitTests.swift#L984-L985

klemenkosir:
What I did is, I got the hexadecimal value of the public key and encoded it into base64. It kinda sucks we don't have support for pems but at least it works. 

Example:
Hex public key
```
9b:e4:16:f3:8b:70:1c:a4:81:a3:bd:12:cb:41:73:f0:2b:ba:1b:f4:9b:1c:b2:b8:d2:50:a9:2b:7a:8f:7d:27
```
remove semicolons and encode to base64.

PR #83, opened 2022-09-16
Add zip header (deflated body) support and documentation

tetraoxygen:
This PR should add support for JWTs with the `"zip": "DEF"` header. 

It does this by using SWCompression, which is unfortunately made necessary by the lack of compression APIs in Foundation on Linux. If anyone has ideas for getting around this, they would be extremely welcome (I considered copying just the necessary files from the dependency and giving appropriate credit, but decided against it for maintainability).

This PR also adds some documentation for `JWTParser`, `JWTSerializer`, `JWTSigner`, and `JWTSigners`.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,13 +16,15 @@ let package = Package(
         MANGLE_END */
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0" ..< "3.0.0")
+        .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0" ..< "3.0.0"),
+		.package(url: "https://github.com/tsolomko/SWCompression.git", from: "4.7.0")
     ],
     targets: [
         .target(name: "CJWTKitBoringSSL"),
         .target(name: "JWTKit", dependencies: [
             .target(name: "CJWTKitBoringSSL"),
             .product(name: "Crypto", package: "swift-crypto"),
+			.product(name: "SWCompression", package: "SWCompression")
         ]),
         .testTarget(name: "JWTKitTests", dependencies: [
             .target(name: "JWTKit"),
--- a/Sources/JWTKit/JWTParser.swift
+++ b/Sources/JWTKit/JWTParser.swift
@@ -28,14 +28,14 @@ struct JWTParser {
     func payload<Payload>(as payload: Payload.Type) throws -> Payload
         where Payload: JWTPayload
     {
-		var decodedPayload = Data(self.encodedPayload.base64URLDecodedBytes())
+        var decodedPayload = Data(self.encodedPayload.base64URLDecodedBytes())
 
-		if let compressionType = try self.header().zip {
-			guard compressionType == "DEF" else {
-				throw JWTError.invalidCompression
-			}
-			decodedPayload = try Deflate.decompress(data: decodedPayload)
-		}
+        if let compressionType = try self.header().zip {
+            guard compressionType == "DEF" else {
+                throw JWTError.invalidCompression
+            }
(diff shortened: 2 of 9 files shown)


Package.swift
```diff
@@ -16,13 +16,15 @@ let package = Package(
         MANGLE_END */
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0" ..< "3.0.0")
+        .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0" ..< "3.0.0"),
+		.package(url: "https://github.com/tsolomko/SWCompression.git", from: "4.7.0")
```
Joannis, line 20: I'd prefer a dependency that's closer to the community standards such as [Adam Fowler's](https://github.com/adam-fowler/compress-nio/blob/main/Package.swift) gzip/deflate support. Adam is a member of the SSWG, which makes me trust the maintenance of the package more. Since Vapor/JWT-Kit is a critical and widely used library, I'm not keen on adding dependencies.

Is there a way we could add compression support using a protocol - leaving the compression algorithms themselves up to other packages? That way we wouldn't have to introduce new dependencies.

NIT: Spacing

Package.swift
```diff
@@ -16,13 +16,15 @@ let package = Package(
...
     ],
     targets: [
         .target(name: "CJWTKitBoringSSL"),
         .target(name: "JWTKit", dependencies: [
             .target(name: "CJWTKitBoringSSL"),
             .product(name: "Crypto", package: "swift-crypto"),
+			.product(name: "SWCompression", package: "SWCompression")
```
Joannis, line 27: NIT: Spacing

Sources/JWTKit/JWTParser.swift
```diff
@@ -1,13 +1,17 @@
 import class Foundation.JSONDecoder
+import struct Foundation.Data
```
Joannis, line 2: I assume we need Foundation.Data for the imported library right?

Tests/JWTKitTests/JWTKitTests.swift
```diff
@@ -643,6 +643,25 @@ class JWTKitTests: XCTestCase {
...
+
+        let jwt = try JWTSigner.es256(
+            key: .private(pem: ecdsaPrivateKey)
+        ).sign(test, zip: .deflate)
+
+        let payload = try JWTSigner.es256(
+            key: .public(pem: ecdsaPublicKey)
+        ).verify(jwt, as: TestPayload.self)
+
+        XCTAssertEqual(test, payload)
```
Joannis, line 662: We should also compare the expected behaviour against known strings. Such as a precompressed JWT (no valdiation perhaps) that we check for correct decompression. I think that should happen for each algorithm, too.

Joannis (changes requested):
Sorry for the delays, I think it's wise to ping us on Discord if your PR is hanging.

Package.swift
gwynne: What @Joannis said - I'm not comfortable requiring a dependency that's not either one of the Apple packages or part of the SSWG collection.

That being said, while a protocol is a good idea, it would also be desirable to provide a default implementation (based upon something that fits the aforementioned criteria for a "safe" dependency).

tetraoxygen: Oh, I hadn’t seen that library at the time, I’ll go take a look! I’d advocate for Gwynne’s approach here, since deflate is really the only compressed JWT type in use at the moment. If adding another dependency (even a safe one) is problematic for this, that could be something like a JWTCompression library with like 10 lines.

ptoffy:
Closing this as with version 5 such functionality can be implemented with custom headers and custom serialising/parsing

PR #84, opened 2022-10-30
JWS/x5c support with ES256 keys

WFT:
Verify JWS tokens using x5c header parameters (ES256 keys only right now).

This PR is based on @0xTim’s work in #75, but uses the X509_STORE/X509_verify_cert API rather than re-implementing that logic in Swift. 

Details on x5c: https://www.rfc-editor.org/rfc/rfc7515#section-4.1.6


WFT:
I rewrote the shape of the API slightly to be its own class so that the root certs could be parsed once at application startup & then re-used for each request, rather than reading the certificates over and over again since that seemed like unnecessary overhead. I included some type-safe wrapper structs for various BoringSSL pointers because I found it easier to read & easier to make sure I wasn’t confusing pointers. Only ES256 is supported because that was what the original PR supported.

My first commit (this PR also has 0xTim’s commits) leaves the public API the same from the original PR. The final commit replaces it with the X5CVerifier class.

Open questions:
- Am I using the C API right? I think so, but there’s not a lot of *simple* sample code out there.
	- Memory safety: I’ve passed the tests with ASAN, so I think there’s no use-after-free issues in here. I don’t think I’m leaking any memory here either, but I haven’t tested.
	- Thread safety: my understanding is that with BoringSSL there’s no need to specially configure thread safety & that since I’m creating the X509_STORE on a single thread and never mutating it again, that should be just fine. But honestly I’m a bit confused about what the rules are. It passed tests with `swift test --sanitize=thread --parallel --num-workers=8`.
- Tests:
	- I’m pretty sure I built the certificates used for testing correctly, but this is my first time doing it!
	- Should I also commit the certificates, keys, and python script I used for generating test cases?
	- Any other cases I should consider?

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+    ///
+    /// - Parameters:
+    ///   - token: The JWS to verify.
+    ///   - payload: The type to decode from the token payload.
+    /// - Returns: The decoded payload, if verified.
+    public func verifyJWS<Payload: JWTPayload>(
+        _ token: String,
+        as payload: Payload.Type = Payload.self
+    ) throws -> Payload {
+        try self.verifyJWS(Array(token.utf8), as: payload)
```
gwynne, line 61:
```suggestion
        try self.verifyJWS(Array(token.utf8), as: Payload.self)
```

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+    {
+        let parser = try JWTParser(token: token)
+        let header = try parser.header()
+        guard let x5c = header.x5c, !x5c.isEmpty else {
+            throw JWTError.generic(identifier: "JWS", reason: "No x5c certificates provided")
+        }
+
+        guard header.alg == "ES256" else {
+            throw JWTError.generic(identifier: "JWS", reason: "Only ES256 is currently supported")
+        }
```
gwynne, line 85:
```suggestion

        guard header.alg == "ES256" else {
            throw JWTError.generic(identifier: "JWS", reason: "Only ES256 is currently supported")
        }
        guard let x5c = header.x5c, !x5c.isEmpty else {
            throw JWTError.generic(identifier: "JWS", reason: "No x5c certificates provided")
        }

```

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+    func freeAll() {
+        // For future maintainers:
+        // There's also a `CJWTKitBoringSSL_sk_X509_free`
+        // that just frees the stack and not the X509s
+        // inside. That may be helpful to wrap.
+        CJWTKitBoringSSL_sk_X509_pop_free(value, CJWTKitBoringSSL_X509_free)
+    }
+
+    func push(_ x509: X509Pointer) throws {
+        if CJWTKitBoringSSL_sk_X509_push(value, x509.value) == 0 {
```
gwynne, line 219:
```suggestion
        guard CJWTKitBoringSSL_sk_X509_push(value, x509.value) != 0 else {
```

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+
+    /// Get the X509 at the index (0-based).
+    subscript(index: Int) -> X509Pointer? {
+        if let pointer = CJWTKitBoringSSL_sk_X509_value(value, index) {
+            return X509Pointer(value: pointer)
+        } else {
+            return nil
+        }
```
gwynne, line 230:
```suggestion
        return CJWTKitBoringSSL_sk_X509_value(value, index).map(X509Pointer.init(value:))
```

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    func free() {
+        CJWTKitBoringSSL_X509_STORE_CTX_free(value)
+    }
+
+    /// Verifies the context as currently built.
+    func verify() throws {
+        if CJWTKitBoringSSL_X509_verify_cert(value) != 1 {
```
gwynne, line 266:
```suggestion
        guard CJWTKitBoringSSL_X509_verify_cert(value) == 1 else {
```

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+private struct X509TrustStore {
+    var value: OpaquePointer
+
+    init() throws {
+        let trustedStore = CJWTKitBoringSSL_X509_STORE_new()
+        if let trustedStore = trustedStore {
+            value = trustedStore
+        } else {
+            throw JWTError.generic(identifier: "JWS", reason: "OpenSSL failure")
+        }
```
gwynne, line 176:
```suggestion
        guard let trustedStore = CJWTKitBoringSSL_X509_STORE_new() else {
            throw JWTError.generic(identifier: "JWS", reason: "OpenSSL failure")
        }
        self.value = trustedStore
```

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /// Add this certificate to the trust store.
+    func trust(_ cert: X509Pointer) throws {
+        if  CJWTKitBoringSSL_X509_STORE_add_cert(value, cert.value) != 1 {
```
gwynne, line 181:
```suggestion
        guard  CJWTKitBoringSSL_X509_STORE_add_cert(value, cert.value) == 1 else {
```

gwynne (commented):
I'm not fully qualified to judge the correctness of the BoringSSL API usage, but here are some pedantic style nits (well, pedantic _and_ chosen to improve clarity and match existing code style, anyhow 😅)

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        // Setup an untrusted chain using all the certificates in the x5c.
+        let untrustedChain = try X509Chain()
+        defer { untrustedChain.freeAll() }
+
+        for cert in x5c {
+            let x509 = try X509Pointer(x5c: cert)
+            try untrustedChain.push(x509)
+        }
```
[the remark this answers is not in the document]
WFT: This has a slightly different effect than the original code — it won't free the chain if pushing one of the certs fails. So I think we either shouldn't take this suggestion OR we should fix that memory leak in this suggestion. I can't really come up with a more clear phrasing than the original, but happy to take more suggestions!

gwynne:
Oh, last but not least,

@Lukasa Hey Cory, bring that expert eye of yours over here from the other PR 😉

WFT:
Fixed a type error introduced by the accepted suggestions. Tests pass for me locally again. 

I don't know what the "test exports" failure is about; happy to chase that down if you can give me a pointer @gwynne

Sources/JWTKit/X5CVerifier.swift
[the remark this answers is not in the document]
gwynne: Oops, you're right. Darn order of operations... there's a reason not everybody likes my "clever" ways of using the stdlib algorithms! 😂 I'd say leave it as it is and ignore my suggestion in this case 🙂

gwynne:
With `main` merged in, the issue with the exports test is that test is specifically intended to test that the package still compiles when all uses of `@_exported import` are suppressed (we should never have used `@_exported` at all at all, but we can't just remove them outright because re-exported modules become part of the public API of the importing module); DoCC (understandably) doesn't handle re-exported modules terribly well, hence this check.

Fortunately the solution is simple - add the missing individual imports to each affected individual file. It looks like all you need to do is add `import Foundation` to the two new files, if I'm reading correctly. To check your build locally, comment out the three imports in `Sources/Utilities/Exports.swift` and try building (make sure to revert the file before committing!).

WFT:
Thanks @gwynne - added the missing imports & tested locally as you suggested.

gwynne (approved):

PR #85, opened 2022-11-22
Add support to PSS algorithm family

MFranceschi6:
Hi all

This pull request adds the support to `PS256`, `PS384` and `PS512` algorithms to this already wonderful jws library.

It doesn't have any impact on the public APIs. It just adds 3 new `JWTSigners` and it gives support for the 3 algorithms in the `JWKSigner` out of the box.

For the cryptographic part, it relies on `BoringSSL` existing functions

We also have added tests to verify the correct implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ JWTKit provides APIs for signing and verifying JSON Web Tokens ([RFC7519](https:
 
 - Verifying (parsing)
 - Signing (serializing)
-- RSA (RS256, RS384, RS512)
+- RSA (RS256, RS384, RS512, PS256, PS384, PS512)
 - ECDSA (ES256, ES384, ES512)
 - HMAC (HS256, HS384, HS512)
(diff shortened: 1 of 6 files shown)


ptoffy:
Hi @MFranceschi6, thanks for the PR and sorry it took a while to respond

We're currently working on a JWT rewrite over [here](https://github.com/vapor/jwt-kit/pull/107) to switch to [swift-crypto](https://github.com/apple/swift-crypto) rather than using BoringSSL. The API switch is actually already complete and we're trying to catch up all with the issues we can before releasing v5. Would you be up to redoing this PR using swift-crypto instead?

MFranceschi6:
Hi there, sure. Let me close this PR and redo it into the `jwtkit-5` branch

ptoffy:
@MFranceschi6 actually you might want to wait for #111 to close as that brings some structural changes

PR #91, opened 2023-04-20
Add option to use a custom JSONDecoder for each JWTPayload type

betaphi:
Introducing the `jsonDecoder()` static method for `JWTPayload`, which in turn gets used by `JWTParser` when decoding the payload.

This is needed because Apple StoreKit JWT payload dates are encoded in `millisecondsSince1970` format for example. See documentation of [JWSTransactionDecodedPayload](https://developer.apple.com/documentation/appstoreservernotifications/jwstransactiondecodedpayload) for example. 

The implementation that I have chosen is a non-breaking change as I have created a default implementation of the `jsonDecoder()` method in an extension of `JWTPayload`.

### DIFF
--- a/Sources/JWTKit/JWTParser.swift
+++ b/Sources/JWTKit/JWTParser.swift
@@ -19,14 +19,17 @@ struct JWTParser {
     }
 
     func header() throws -> JWTHeader {
-        try self.jsonDecoder()
+        let jsonDecoder = JSONDecoder()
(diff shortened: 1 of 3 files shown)


0xTim:
@betaphi thanks for this! We really appreciate the contribution! I'm going to close this in favour of #95 as that provides both custom encoders and decoders

PR #92, opened 2023-05-02
AppleIdentityToken: Adds `real_user_status` claim

### LINKED ISSUE #88: JWTKit misses Apple's `real_user_status`
Apple has a `real_user_status` field in their tokens.

> This claim is present only in iOS 14 and later, macOS 11 and later, watchOS 7 and later, tvOS 14 and later. The claim isn’t present or supported for web-based apps.

https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple

**Describe the solution you'd like**
Implement the `real_user_status` field as an optional in Apple tokens.

qalandarov:
Adds the missing `real_user_status` claim which helps detecting if the user is likely a real person. Closes #88

### DIFF
--- a/Sources/JWTKit/Vendor/AppleIdentityToken.swift
+++ b/Sources/JWTKit/Vendor/AppleIdentityToken.swift
@@ -1,17 +1,6 @@
 /// - See Also:
 /// [Retrieve the User’s Information from Apple ID Servers](https://developer.apple.com/documentation/signinwithapplerestapi/authenticating_users_with_sign_in_with_apple)
 public struct AppleIdentityToken: JWTPayload {
-
-    /// Taken from https://developer.apple.com/documentation/authenticationservices/asuserdetectionstatus
-    public enum UserDetectionStatus : Int, Codable {
-        /// Not supported on current platform, ignore the value
-        case unsupported
-        /// We could not determine the value.  New users in the ecosystem will get this value as well, so you should not block these users, but instead treat them as any new user through standard email sign up flows
-        case unknown
-        /// A hint that we have high confidence that the user is real
-        case likelyReal
-    }
-
     enum CodingKeys: String, CodingKey {
         case nonce, email
         case issuer = "iss"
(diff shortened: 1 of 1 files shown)


Sources/JWTKit/Vendor/AppleIdentityToken.swift
```diff
@@ -1,6 +1,17 @@
 /// - See Also:
 /// [Retrieve the User’s Information from Apple ID Servers](https://developer.apple.com/documentation/signinwithapplerestapi/authenticating_users_with_sign_in_with_apple)
 public struct AppleIdentityToken: JWTPayload {
+
+    /// Taken from https://developer.apple.com/documentation/authenticationservices/asuserdetectionstatus
+    public enum UserDetectionStatus : Int, Codable {
```
0xTim, line 6: I'm a little hesitant to introduce a new public enum here, especially to model something we don't have control over. The reason being that if Apple release a new status type, that's technically a break change for us if we add the new case to the enum. 

The best way around this would be to keep the public enum, but just make each case a `static let` property so adding new properties is non-breaking. We can then use an internal enum for decoding and storage, as adding new cases to that is non-breaking and then map between the enum cases and static properties. Does that makes sense?

0xTim (changes requested):

Sources/JWTKit/Vendor/AppleIdentityToken.swift
qalandarov: Conceptually I got it, but I haven't done it before that way, so is this something that you're asking @0xTim?

```
public enum UserDetectionStatus: Int, Codable {
    // Static let properties for each case
    public static let unsupported = UserDetectionStatus(rawValue: 0)
    public static let unknown = UserDetectionStatus(rawValue: 1)
    public static let likelyReal = UserDetectionStatus(rawValue: 2)

    // Internal enum for decoding and storage
    internal enum RawValue: Int, Codable {
        case unsupported
        case unknown
        case likelyReal
    }

    // Mapping between the enum cases and static properties
    public init(from decoder: Decoder) throws {
        let rawValue = try decoder.singleValueContainer().decode(RawValue.self)
        switch rawValue {
        case .unsupported:
            self = UserDetectionStatus.unsupported
        case .unknown:
            self = UserDetectionStatus.unknown
        case .likelyReal:
            self = UserDetectionStatus.likelyReal
        }
    }

    public func encode(to encoder: Encoder) throws {
        var container = encoder.singleValueContainer()
        switch self {
        case .unsupported:
            try container.encode(RawValue.unsupported)
        case .unknown:
            try container.encode(RawValue.unknown)
        case .likelyReal:
            try container.encode(RawValue.likelyReal)
        }
    }
}
```

0xTim: @qalandarov yep exactly!

qalandarov: Had to make slight updates compared to the above example

0xTim (approved):
Thanks @qalandarov ! Can you merge in the latest changes and we can get this released

PR #93, opened 2023-06-02
Update the vendored BoringSSL

gwynne:
Update the vendored BoringSSL to the latest. This also brings in the latest updates to the vendoring scripts themselves from upstream, updates CI (what? In a PR from Gwynne? Say it ain't so!), and bumps the minimum Swift version to 5.6 to be in line with Vapor.


Joannis (approved):
Looks good, but I think a lot of these changes are for windows support. Shouldn't we add (non-required) CI for that?

gwynne:
CI for Windows is pointless until SwiftPM is more stable and NIO works. Besides, we're not updating to get changes for Windows support anyway; we just want to stay up to date with bugfixes etc. in general (also it's been quite awhile since the vendored copy was last bumped).

PR #94, opened 2023-06-15
Add optional org_id field to AppleIdentityToken

RanLearns:
### **Motivation and Reasoning**

When working with Managed Apple IDs (schools and business) there can be a need to decode an additional field of a user's org_id when they sign in.

### DIFF
--- a/Sources/JWTKit/Vendor/AppleIdentityToken.swift
+++ b/Sources/JWTKit/Vendor/AppleIdentityToken.swift
@@ -41,7 +41,7 @@ public struct AppleIdentityToken: JWTPayload {
     /// The user's email address.
     public let email: String?
 
-    /// Managed Apple ID organization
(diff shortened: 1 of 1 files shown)


Sources/JWTKit/Vendor/AppleIdentityToken.swift
```diff
@@ -40,6 +41,9 @@ public struct AppleIdentityToken: JWTPayload {
     /// The user's email address.
     public let email: String?
 
+    /// Managed Apple ID organization
```
gwynne, line 44:
```suggestion
    /// Managed Apple ID organization (see https://developer.apple.com/documentation/rosterapi/integrating_with_roster_api_and_sign_in_with_apple)
```

gwynne (approved):
LGTM!

PR #95, opened 2023-07-17
Add customizable JSON encoders and decoders

gwynne:
Add new, fully source-compatible APIs to `JWTSigners` and `JWTSigner` which allow specifying custom JSON encoders and decoders. Also provides the `JWTJSONEncoder` and `JWTJSONDecoder`protocols, which allow using alternative JSON implementations.

Custom coders specified for a single `JWTSigner` affect token parsing and signing performed only by that signer. Custom coders specified on a `JWTSigners` object will become the default coders for all signers added to that object, unless a given signer already specifies its own custom coders.

The default encoding and decoding implementation provided for `JWTUnixEpochClaim` (of which `ExpirationClaim` (`exp`), `IssuedAtClaim` (`iat`), and `NotBeforeClaim` (`nbf`) are examples) has been adjusted to encode and decode its `Date` value directly, rather than performing the explicit conversion to and from a `Double`. This allows these claims to take advantage of the `dateEncodingStrategy` and `dateDecodingStrategy` specified on custom JSON coders. (It also gives a bit of the lie to the name `JWTUnixEpochClaim`, but it's public API, so I left it alone.)

The default coders in use remain the same: An encoder and decoder which use the `.secondsSince1970` date encoding/decoding strategy. As such, neither the new support nor the change to `Date`-based claims affects how tokens are signed or verified unless custom coders with different strategies are specified (that being, after all, the original need which inspired these changes).

Finally, an `.integerSecondsSince1970` date encoding and decoding strategy has been added to the public API for the benefit of interoperation with JWT implementations - such as that used by GitHub - which require the aforementioned date-based claims to specify values as an integer number of seconds. (As GitHub proves, while this is in violation of the definition of `NumericDate` given by RFC 7519 § 2, which explicitly permits floating-point values, it nonetheless appears in the wild.)

This is a `semver-minor` release, as it adds new public API (although it has been careful to fully retain source compatibility, at the cost of a goodly amount of silly-looking repetition in the implementation - please, do _not_ ask me if I know what default parameter values are! 😂).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         MANGLE_END */
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-crypto.git", "1.0.0" ..< "3.0.0")
+        .package(url: "https://github.com/apple/swift-crypto.git", from: "2.0.0")
     ],
     targets: [
         .target(name: "CJWTKitBoringSSL"),
--- a/README.md
+++ b/README.md
@@ -1,29 +1,28 @@
-<img 
-    src="https://user-images.githubusercontent.com/1342803/59471117-1c77b300-8e08-11e9-838e-441b280855b3.png" 
-    height="64" 
-    alt="JWTKit"
-/>
+<p align="center">
+    <img 
+        src="https://user-images.githubusercontent.com/1342803/59471117-1c77b300-8e08-11e9-838e-441b280855b3.png" 
+        height="64" 
+        alt="JWTKit">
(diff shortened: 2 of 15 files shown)


MahdiBM (approved):
LGTM

MahdiBM (approved):
I need to ask ChatGPT for synonyms for LGTM 😅

0xTim (approved):
Looks good to me. The tests are a bit thin but then always have been.

Next step - removing BoringSSL 👀

PR #98, opened 2023-10-04
Increase permitted range for crypto-kit dependency

AndyHeardApps:
Currently the crypto-kit package has a permitted version range of `2.0.0 ..< 3.0.0`. With the release of crypto-kit `3.0.0`, this range is preventing Vapor apps from making use of the latest versions. Increasing the range to `2.0.0 ..< 4.0.0` allows apps to use the latest release without forcing an update.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
         MANGLE_END */
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


0xTim (approved):
Thanks!

PR #99, opened 2023-10-09
Move away from BoringSSL

ptoffy:
This aims to remove most if not all usages of BoringSSL and replace it with [swift-crypto](https://github.com/apple/swift-crypto/) and [swift-certificates](https://github.com/apple/swift-certificates)


Tests/JWTKitTests/X5CTests.swift
```diff
@@ -155,9 +161,36 @@ final class X5CTests: XCTestCase {
...
 
-    func verify(using signer: JWTSigner) throws {
+    func verify(using _: JWTSigner) throws {
         if !cool {
             throw JWTError.claimVerificationFailure(name: "cool", reason: "not cool")
         }
     }
 }
+
+func XCTAssertNoThrowAsync<T>(
```
0xTim, line 171: These should be in a separate file

Tests/JWTKitTests/JWTKitTests.swift
```diff
@@ -1,6 +1,6 @@
 import XCTest
 #if os(Linux)
-import FoundationNetworking
+    import FoundationNetworking
```
0xTim, line 3: Do we need this

Package.swift
```diff
@@ -1,32 +1,41 @@
...
     ],
     products: [
         .library(name: "JWTKit", targets: ["JWTKit"]),
-        /* This target is used only for symbol mangling. It's added and removed automatically because it emits build warnings. MANGLE_START
-        .library(name: "CJWTKitBoringSSL", type: .static, targets: ["CJWTKitBoringSSL"]),
-        MANGLE_END */
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-crypto.git", "2.0.0" ..< "4.0.0")
+        .package(url: "https://github.com/apple/swift-crypto.git", "2.6.0" ..< "4.0.0"),
```
0xTim, line 16: Let's just do 3.X.X for now

Package.swift
```diff
@@ -1,32 +1,41 @@
...
+            dependencies: [
+                "JWTKit",
+            ],
+            resources: [
+                .copy("TestVectors"),
+            ]
+        ),
     ],
-     cxxLanguageStandard: .cxx11
+    cxxLanguageStandard: .cxx11
```
0xTim, line 40: I don't think we need this anymore

0xTim (commented):
The code changes in the actual business logic that go from hundreds of lines of horrible unsafe code to a few nice lines of Swift are amazing

Initial comments

Sources/JWTKit/Utilities/Base64URL.swift
```diff
@@ -1,25 +1,46 @@
...
-        for (i, byte) in self.enumerated() {
+        for (i, byte) in enumerated() {
             switch byte {
-            case 0x2D: self[self.index(self.startIndex, offsetBy: i)] = 0x2B
-            case 0x5F: self[self.index(self.startIndex, offsetBy: i)] = 0x2F
+            case 0x2D: self[index(startIndex, offsetBy: i)] = 0x2B
+            case 0x5F: self[index(startIndex, offsetBy: i)] = 0x2F
             default: break
             }
         }
```
gwynne, line 46:
```swift
        for idx in self.indices {
            switch self[idx] {
            case 0x2d/* - */: self[idx] = 0x2b/* + */
            case 0x5f/* _ */: self[idx] = 0x2f/* / */
            default: break
            }
        }
```
Not only more readable but HUGELY more efficient...

Sources/JWTKit/Utilities/Base64URL.swift
```diff
@@ -36,8 +57,8 @@ private extension Data {
     mutating func base64URLEscape() {
         for (i, byte) in enumerated() {
             switch byte {
-            case 0x2B: self[self.index(self.startIndex, offsetBy: i)] = 0x2D
-            case 0x2F: self[self.index(self.startIndex, offsetBy: i)] = 0x5F
+            case 0x2B: self[index(startIndex, offsetBy: i)] = 0x2D
+            case 0x2F: self[index(startIndex, offsetBy: i)] = 0x5F
             default: break
             }
         }
```
gwynne, line 64:
```swift
        for idx in self.indices {
            switch self[idx] {
            case 0x2b/* + */: self[idx] = 0x2d/* - */
            case 0x2f/* / */: self[idx] = 0x5f/* _ */
            default: break
            }
        }
```
Not only more readable but HUGELY more efficient...

gwynne (changes requested):
Bunches of nits all hither and thither

0xTim (approved):
There's probably a bit more work we can do on docs, like ensuring all types are linked, everything is public, a basic usage index page etc etc but that can all be done later.

BoringSSL removed! 🥳

PR #103, opened 2023-10-09
Fix typo for `JWTError.signatureVerificationFailed` case

davdroman:
Succeeds #61 cause I had to re-fork in order to do further edits. Hope that's alright.


gwynne (approved):

0xTim:
Isn't the API breakage checker correct here? Anyone doing a switch will have a build error

davdroman:
@0xTim oh yes, of course. Happy to drop it if it's a no-go.

gwynne:
> Isn't the API breakage checker correct here? Anyone doing a switch will have a build error

Ouch, you're right, the compiler will still complain the switch isn't exhaustive 😞

0xTim:
Thanks for this @davdroman, thankfully we're working on a major update at the moment where this is already fixed

PR #111, opened 2023-11-01
Implement `JWTKeyCollection` and hide `JWTSigner`

ptoffy:
This also starts adding `Sendable` support so we might as well implement it here directly


Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -0,0 +1,239 @@
+import Foundation
+
+/// A collection of JWT and JWK signers for handling JSON Web Tokens (JWTs).
+///
+/// This class provides methods to manage multiple signers, allowing the addition and retrieval of ``JWTSigner`` and ``JWKSigner`` instances.
```
0xTim, line 5: Nit this is an actor not class

We should also tidy up the language as we don't expose signers anymore

Sources/JWTKit/JWTKeyCollection.swift
0xTim, line 1: Overall comment about the DocC comments - we should avoid talking about `JWTSigner`s and other non public types in docs that will be what the end user sees to avoid confusion

Sources/JWTKit/Vendor/AppleIdentityToken.swift
```diff
@@ -63,10 +63,10 @@ public struct AppleIdentityToken: JWTPayload {
     }
 }
 
-extension AppleIdentityToken {
+public extension AppleIdentityToken {
     /// Taken from https://developer.apple.com/documentation/authenticationservices/asuserdetectionstatus
     /// With slight modification to make adding new cases non-breaking.
-    public struct UserDetectionStatus: OptionSet, Codable {
+    struct UserDetectionStatus: OptionSet, Codable, Sendable {
```
0xTim, line 69: I think this type needs to be public so users can see it

Tests/JWTKitTests/ECDSATests.swift
```diff
@@ -3,26 +3,27 @@ import BigInt
 import XCTest
 
 final class ECDSATests: XCTestCase {
-    func testECDSADocs() throws {
-        let signers = JWTSigners()
-        XCTAssertNoThrow(try signers.use(.es256(key: .public(pem: ecdsaPublicKey))))
+    func testECDSADocs() async throws {
+        XCTAssertNoThrow(try ES256Key.public(pem: ecdsaPublicKey))
```
0xTim, line 7: Shouldn't this test try adding the key to the collection

Tests/JWTKitTests/ECDSATests.swift
```diff
@@ -113,16 +115,15 @@ final class ECDSATests: XCTestCase {
...
+        let foo = try await keys.verify(jwt, as: Foo.self)
         XCTAssertEqual(foo.bar, 42)
     }
 
-    func testVerifyingECDSAKeyUsingJWKBase64URL() throws {
+    func testVerifyingECDSAKeyUsingJWKBase64URL() async throws {
         struct Foo: JWTPayload {
             var bar: Int
-            func verify(using _: JWTSigner) throws {}
+            func verify(using _: JWTAlgorithm) throws {}
```
0xTim, line 126: Is it worth making verify async on the protocol to align with everything else?

Tests/JWTKitTests/JWTKitTests.swift
```diff
@@ -94,34 +95,33 @@ class JWTKitTests: XCTestCase {
...
-        let jwt = try JWTSigner.hs256(key: "secret".bytes)
-            .sign(ExpirationPayload(exp: exp))
+        let keyCollection = await JWTKeyCollection().addHS256(key: "secret".bytes)
+        let jwt = try await keyCollection.sign(ExpirationPayload(exp: exp))
         let parser = try JWTParser(token: jwt.bytes)
         try XCTAssertEqual(parser.header(jsonDecoder: .defaultForJWT).typ, "JWT")
         try XCTAssertEqual(parser.header(jsonDecoder: .defaultForJWT).alg, "HS256")
         try XCTAssertEqual(parser.payload(as: ExpirationPayload.self, jsonDecoder: .defaultForJWT).exp, exp)
-        try parser.verify(using: .hs256(key: "secret".bytes))
+        try await parser.verify(using: keyCollection.signer()!.algorithm)
```
0xTim, line 114: One comment now I see the API, do we want `signer()` to throw instead of returning an optional? Having the optional here feels unnatural

0xTim (commented):
Initial comments

Sources/JWTKit/Vendor/AppleIdentityToken.swift
ptoffy: It is, SwiftFormat put the `public` on the extension rather than on the struct itself

Tests/JWTKitTests/ECDSATests.swift
ptoffy: Adding a key to the collection doesn't throw

Tests/JWTKitTests/JWTKitTests.swift
ptoffy: There's actually `require(kid:, alg:)` which I added later and wraps the `signer()` into a throwing function. Do you think we can use that or should we just make `signer` throwing and remove `require`?

ptoffy: Just to be clear these aren't public anyways

0xTim: Do you need to test the internals then? Is it possible to rewrite the tests so that we still test the same behaviour but only via the public API (it should be). 

IMO a good test suite combined with a solid API doesn't need to use `@testable`

ptoffy: I think I solved this specific instance but there's still some stuff we need to test which is not be public, like the ECDSA parameters, the `JWTParser` or the `PrimeGenerator`. Do you think we can make this any better?

0xTim: We can add those to a test file dedicated to internal tests. Feel free to create an issue and do it in a separate PR to unblock this

0xTim (approved):
Good to merge, we'll do an API audit before any release anyway

PR #112, opened 2023-11-03
Add RSA-PSS signature algorithm support

MFranceschi6:
Hi all!

I've made some "breaking" changes inside the `JWTSigner+RSA.swift` file.
Hope you will find them useful 😁 

I've removed the `padding` parameter from every method.
The padding is already defined by the name of the algorithm.
Also the RS algorithm family uses the [RSASSA-PKCS1-v1_5](https://datatracker.ietf.org/doc/html/rfc7518#section-3.3) so I've changed the padding for those methods to `.insecurePKCS1v1_5`.

Basically you were signing JWTs with the algorithm name `RS256` but it actually was a `PS256` signature.

You can check this yourself.
From your base branch
Go to `JWTKitTests.swift` file.
Test named `testJWKSigner`
print the `publicKey` and the `data` variables.
Then [jwt.io](https://jwt.io/) to verify the signature.
Expected result ❌ 

Then go into the `JWTSigner+RSA.swift` change the name at line 10 from `RS256` to `PS256`.
Redo the test and see that now jwt.io will say that the signature is correct.
Expected result ✅


ptoffy (changes requested):
That's on me 😅 PSS was going to be the next step but cool that you figured it out first! Happy to accept it, though #35 brought some structural changes which need to be addressed here, mainly around using `JWTKeyCollection` instead of `JWTSigners` which was removed.
Also we'll need some PSS tests too! 😃

gwynne (approved):

PR #116, opened 2023-11-06
Add `Sendable` support

ptoffy:
This aims to add Sendable support to the objects that don't have it already

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -40,6 +40,7 @@ let package = Package(
                 .copy("TestVectors"),
             ],
             swiftSettings: [
+                .enableExperimentalFeature("StrictConcurrency"),
                 .enableUpcomingFeature("ConciseMagicFile"),
             ]
         ),
--- a/Package.swift
+++ b/Package.swift
@@ -18,15 +18,19 @@ let package = Package(
         .package(url: "https://github.com/attaswift/BigInt.git", from: "5.3.0"),
     ],
     targets: [
-        .target(name: "JWTKit", dependencies: [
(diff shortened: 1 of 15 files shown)


Package.swift
```diff
@@ -18,22 +18,29 @@ let package = Package(
...
         .testTarget(
             name: "JWTKitTests",
             dependencies: [
                 "JWTKit",
             ],
             resources: [
                 .copy("TestVectors"),
+            ],
+            swiftSettings: [
+                .enableUpcomingFeature("ConciseMagicFile"),
```
0xTim, line 43: We should add strict Sendable checking to the tests as well just to prepare for Swift 6

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -14,7 +14,7 @@ import X509
 ///
 /// See [RFC 7515](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.6)
 /// for details on the `x5c` header parameter.
-public class X5CVerifier {
+public final class X5CVerifier: Sendable {
```
0xTim, line 17: Could we make this a struct?

0xTim (changes requested):

0xTim:
Cory mentioned that Crypto types as are value types and can be considered Sendable so this PR introduces no issues.

 Added https://github.com/apple/swift-crypto/issues/208 to track Sendable to Crypto Extras as well

0xTim (approved):

PR #117, opened 2023-11-06
Alter X5CVerifier for correct StoreKit 2 receipt and App Store Server Notifications validation

betaphi:
### Motivation

Apple StoreKit 2 receipts (Transaction/RenewalInfo) ([documentation](https://developer.apple.com/documentation/storekit/verificationresult/3868429-jwsrepresentation)) as well as App Store Server Notifications v2 ([documentation](https://developer.apple.com/documentation/appstoreservernotifications/app_store_server_notifications_v2)) require some special treatment in order to correctly verify their signature.

These are:

- As all dates are encoded in milliseconds since 1970, the JSONDecoders must be adjusted accordingly.
- Receipts are stored locally on the users machines and when verified on a server are possibly valid for longer periods of time than the signing certificates that were used by the App Store. Therefore the `signedDate` in the payload of the JWT must be used in order to verify the expiration date of the certificates. For example: on 01.01.2023 the user's subscription is renewed for one year until 01.01.2024. Therefore the transaction's `signedDate` will be 01.01.2023. However, the certificate that is used by Apple to sign the transaction expires on 01.08.2023. Now, with the classical approach, on 01.08.2023 the transaction would not validate anymore because the certificate seems to be expired. In a perfect world, Apple would have already replaced the transaction with a new signature using the latest certificate. But unfortunately this is not how the App Store behaves. Instead, we need to verify the signature using the `signedDate` in the payload of the transaction in order to provide the correct verification result.

### Solution

Add a new public method to `X5CVerifier`, that lets us specify the verificationTime.
In order to prevent the need of a double initialisation of the same `JWTParser` during the verification and later during the actual decoding of the payload, the new method takes a `JWTParser` as a parameter instead of `some DataProtocol` or `String`.

In order to verify such a JWT, we need to go through the following procedure:

1. Create a `JWTParser` with the JWT to be verified.
2. Use the `JWTParser` to decode its payload's `signedDate` value (using a JSONDecoder with the correct `dateDecodingStrategy`)
3. Call the `X5CVerifier`'s new `verifyJWS` method with the `JWTParser` and `validationTime` parameters.

### Testing

I have tested the code locally using live App Store data. However, as this data contains private information I am not able to include the test cases in the PR. Unfortunately, I'm not aware of a way to generate the required JWT data for testing this in a way that does not include private data. Therefore no tests have been added at this point.

The required JWT for testing would have to meet the following criteria:

1. Be X5C signed.
2. One of the leaf certificates should be expired.
3. The payload must contain a key named `signedDate` with a date at which all leaf certificates were valid.
4. The payload must be encoded using the `millisecondsSince1970` date format.

I'd be happy to write the test case should somebody be able to create such a JWT.

### Breaking changes

None. I was careful not to change any public function signatures or behaviours.


0xTim:
Couple of initial comments - this should probably live at the vapor/jwt level rather than JWTKit to match the other Apple specific stuff. I don't particularly want domain specific stuff in the library. Should however definitely have the APIs available to be able to do that if it's not currently possible. 

Second, as probably a moot point due to the point above, subclassing will not work (it's now a struct because Sendable) and as a general rule we do not allow subclassing anywhere in Vapor as it's a cluster hell of hidden and confusing behaviour that should be not allowed in Swift at any level 😅

0xTim:
Re testing, whilst we can't use Apple JWTs for this we should be able to create our own for tests that have the same behaviour (millisecond encoded dates etc)

betaphi:
> Couple of initial comments - this should probably live at the vapor/jwt level rather than JWTKit to match the other Apple specific stuff. I don't particularly want domain specific stuff in the library. Should however definitely have the APIs available to be able to do that if it's not currently possible.

This could certainly be done, however vapor/jwt is - as far as I can tell - currently not up to date with the jwtkit-5 branch here. So unfortunately there is nothing I can do with that regard at this point.
Also, while I understand your intention, I'd like to highlight that vapor/jwt-kit also already includes other "Apple specific stuff".

Finally, I'd like to add that I am using the vapor/jwt-kit library in places where I don't use vapor/jwt because it does not dependent on vapor itself. These use cases would be impossible to cover once this code would move into vapor/jwt instead of stying here.

> Second, as probably a moot point due to the point above, subclassing will not work (it's now a struct because Sendable) and as a general rule we do not allow subclassing anywhere in Vapor as it's a cluster hell of hidden and confusing behaviour that should be not allowed in Swift at any level 😅

While there are certainly points for and against subclassing the question is what the current status of development is. I was under the impression that this was the jwtkit-5 branch but your comment seems to indicate it is rather the sendable branch.

Now here is my proposal:
As the X5C verification and parsing of StoreKit 2 receipts is broken for months now and it seems nobody really cares to get it fixed, I would add the necessary public facing methods here and leave the Apple specific stuff for another time.

betaphi:
@0xTim I have now updated the PR in the hope that this minimal invasive change in code is acceptable for you in order to allow for this huge step in usefulness. After all I would think it could be one of the most prominent use cases for jwtkit to verify App Store receipts.

0xTim:
@betaphi just an FYI - we are discussing the best ways of enabling this behaviour. Apple have actually released a Swift Server library for interacting with the App Store - https://github.com/apple/app-store-server-library-swift

The issue is that they have to run custom verifiers and manual decode and verify the JWTs. This is something I want to fix in JWTKit so we're trying to ensure we have a nice API to do this given we're able to have a brand new API

ptoffy:
Hi @betaphi, the behaviour you're requesting was implemented in #119. [Here](https://github.com/vapor/jwt-kit/blob/jwtkit-5/Tests/JWTKitTests/X5CTests.swift#L119) is a test demonstrating it, so I'll close this PR. Feel free to open an issue or ask on the [Vapor Discord](https://discord.gg/VWUJz6gd) if you have any questions!

PR #119, opened 2023-11-09
Add support for custom time validation X5Cs

ptoffy:
This adds support for X5C chains to verify using custom time settings, which allows for app store receipts verification


Sources/JWTKit/Utilities/CustomizedJSONCoders.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 
 public protocol JWTJSONDecoder: Sendable {
     func decode<T: Decodable>(_: T.Type, from string: Data) throws -> T
+    var dateDecodingStrategy: JSONDecoder.DateDecodingStrategy { get set }
```
0xTim, line 5: We should add this to the encoder as well

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -121,13 +121,27 @@ public struct X5CVerifier: Sendable {
...
 
-        // Setup an untrusted chain using all the certificates in the x5c.
+        // Setup an untrusted chain using all the certificates in the x5c
         let untrustedChain = try CertificateStore(certificateData.map {
             try Certificate(derEncoded: [UInt8]($0))
         })
 
+        var date = Date()
+        var decoder = jsonDecoder
+        if Payload.self is ValidationTimePayload.Type {
```
0xTim, line 131: Couple of lines of comments of why this is required

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -121,13 +121,27 @@ public struct X5CVerifier: Sendable {
             return data
         }
 
-        // Setup an untrusted chain using all the certificates in the x5c.
+        // Setup an untrusted chain using all the certificates in the x5c
         let untrustedChain = try CertificateStore(certificateData.map {
             try Certificate(derEncoded: [UInt8]($0))
         })
 
+        var date = Date()
```
0xTim, line 129: Set as a let and use an else block down below instead of overwriting it if we need to

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -121,13 +121,27 @@ public struct X5CVerifier: Sendable {
...
+        var decoder = jsonDecoder
+        if Payload.self is ValidationTimePayload.Type {
+            decoder.dateDecodingStrategy = .millisecondsSince1970
+        }
+
+        let payload = try parser.payload(as: Payload.self, jsonDecoder: decoder)
+
+        if let validationTimePayload = payload as? ValidationTimePayload {
```
0xTim, line 137: I wonder if we could check the against the type itself rather than the struct to avoid a double parse

Sources/JWTKit/X5CVerifier.swift
```diff
@@ -139,10 +153,14 @@ public struct X5CVerifier: Sendable {
...
-        // Assuming the chain is valid, verify the token was signed by the valid certificate.
+        // Assuming the chain is valid, verify the token was signed by the valid certificate
         let ecdsaKey = try ES256Key.certificate(pem: leafCertificate.serializeAsPEM().pemString)
 
         let signer = JWTSigner(algorithm: ECDSASigner(key: ecdsaKey, algorithm: .sha256, name: headerAlg))
         return try await signer.verify(parser: parser)
     }
 }
+
+public protocol ValidationTimePayload: JWTPayload {
```
0xTim, line 164: Separate file and doc comments

Tests/JWTKitTests/JWTKitTests.swift
```diff
@@ -357,6 +358,45 @@ class JWTKitTests: XCTestCase {
...
+        // MARK: - DataClass
+
+        struct DataClass: Codable {
+            let appAppleId: Int
+            let environment, bundleId: String
+        }
+
+        let verifier = try X5CVerifier(rootCertificates: [cert])
+
+        await XCTAssertNoThrowAsync(try await verifier.verifyJWS(data, as: StoreKitPayload.self))
```
0xTim, line 398: Let's have an assertion on the app ID and environment etc

0xTim (changes requested):

Tests/JWTKitTests/X5CTests.swift
```diff
@@ -119,6 +119,59 @@ final class X5CTests: XCTestCase {
...
+        jsonDecoder.dateDecodingStrategy = .millisecondsSince1970
+
+        var payload: StoreKitPayload?
+        do {
+            payload = try await verifier.verifyJWS(data, as: StoreKitPayload.self, jsonDecoder: jsonDecoder)
+        } catch {
+            XCTFail("Failed with error: \(error.localizedDescription)")
+        }
+
+        if let payload = payload {
```
0xTim, line 168: You can use `XCTUnwrap` for this - also this is 5.9 so it should just be `if let payload`!

Sources/JWTKit/X5C/X5CVerifier.swift
```diff
@@ -121,13 +121,26 @@ public struct X5CVerifier: Sendable {
...
         let untrustedChain = try CertificateStore(certificateData.map {
             try Certificate(derEncoded: [UInt8]($0))
         })
 
+        let payload = try parser.payload(as: Payload.self, jsonDecoder: jsonDecoder)
+
+        let date: Date
+        // Some JWT implementations have the sign date in the payload.
+        // If it's such a payload, we'll use that date for validation
+        if let validationTimePayload = payload as? ValidationTimePayload {
```
0xTim, line 134: I'd still like to see if we can compare against `Payload.Type` here

0xTim (changes requested):
Looking good

Sources/JWTKit/X5C/X5CVerifier.swift
ptoffy: @0xTim I mean we can but to access the property afterwards we still need to cast it

0xTim: Ahh of course

0xTim (approved):

PR #120, opened 2023-11-09
Add ECDSAKey PEM export

ptoffy:


Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -11,7 +11,7 @@ import X509
 ///
 /// - Generic Parameter Curve: The curve type associated with the ECDSA key.
 public struct ECDSAKey<Curve>: ECDSAKeyType where Curve: ECDSACurveType {
-    /// The elliptic curve used by this key.
```
gwynne, line 14: Why this change? It doesn't match style (in fact it does the opposite).

Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -105,6 +105,28 @@ public struct ECDSAKey<Curve>: ECDSAKeyType where Curve: ECDSACurveType {
         return try self.private(pem: string)
     }
 
+    /// Exports the current public key as a PEM encoded string.
+    ///
+    /// - Throws: If the key is not a public key.
```
gwynne, line 110: Your logic, both here and down in `.init(privateKey:publicKey:)` does not account for the fact that a public key can always be derived from a private key (whereas the logic in `generate()` _does_ account for this).

gwynne (changes requested):

gwynne (approved):

PR #125, opened 2023-11-20
Add support for the b64 header

ptoffy:


gwynne:
Per [RFC 7797 § 7](https://www.rfc-editor.org/rfc/rfc7797#section-7), I'm not okay with adding this without a strong use case for it.

ptoffy:
Closing this as we now (#130) support custom headers and custom serialiser/parsers so something like a b64 can be easily implemented as shown [here](https://github.com/vapor/jwt-kit/blob/jwtkit-5/Tests/JWTKitTests/JWTKitTests.swift#L342-L384)

PR #126, opened 2023-11-23
Add option to sign tokens with x5c chains

ptoffy:
This adds a new `x5c` parameter to the sign method, which accepts an array of pem encoded certificates which form an x5c chain. This chain gets added to the JWT's header with the `x5c` field. For the token to be valid, it must be signed with the leaf certificate's private key

### DIFF
--- a/Sources/JWTKit/JWTSerializer.swift
+++ b/Sources/JWTKit/JWTSerializer.swift
@@ -1,3 +1,6 @@
+import Foundation
+import X509
+
 struct JWTSerializer {
     func sign(
         _ payload: some JWTPayload,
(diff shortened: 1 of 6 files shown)


gwynne (approved):

Sources/JWTKit/JWTSerializer.swift
```diff
@@ -1,19 +1,32 @@
...
         // encode header, copying header struct to mutate alg
         var header = JWTHeader()
         header.kid = kid
         header.typ = typ
         header.cty = cty
         header.alg = signer.algorithm.name
 
+        if let x5c, !x5c.isEmpty {
+            let verifier = try X5CVerifier(rootCertificates: [x5c[0]])
+            try await verifier.verifyChain(certificates: x5c)
```
0xTim, line 23: We may want to add a flag to disable verification, for testing etc. But I'm not sure if this is explicitly disallowed

Sources/JWTKit/X5C/X5CVerifier.swift
```diff
@@ -38,6 +38,24 @@ public struct X5CVerifier: Sendable {
         })
     }
 
+    /// Verify a chain of certificates against the trusted root certificates.
+    ///
+    /// - Parameter certificates: The certificates to verify.
+    /// - Throws: A `JWTError` if the chain is invalid.
+    package func verifyChain(certificates: [String]) async throws {
```
0xTim, line 45: Does this need to be package protected instead of internal?

Tests/JWTKitTests/X5CTests.swift
```diff
@@ -165,6 +166,48 @@ final class X5CTests: XCTestCase {
         XCTAssertEqual(data.appAppleId, 1234)
         XCTAssertEqual(data.environment, "Sandbox")
     }
+
+    func testSigningWithX5CChain() async throws {
```
0xTim, line 170: We should test a chain with a cert that isn't verified by the node above to make sure it throws an error

0xTim (changes requested):
Looking good, a couple of questions and more tests needed

0xTim (approved):

PR #127, opened 2023-11-25
Add option to fetch RSA primitives

ptoffy:


Sources/JWTKit/RSA/SubjectPublicKeyInfo.swift
```diff
@@ -0,0 +1,106 @@
...
+extension RFC5480AlgorithmIdentifier {
+    static let rsa256 = RFC5480AlgorithmIdentifier(
+        algorithm: .AlgorithmIdentifier.sha256WithRSAEncryption
+    )
+
+    static let rsa384 = RFC5480AlgorithmIdentifier(
+        algorithm: .AlgorithmIdentifier.sha384WithRSAEncryption
+    )
+
+    static let rsa521 = RFC5480AlgorithmIdentifier(
```
MahdiBM, line 103: 521 😱

gwynne: Not a typo.

MahdiBM: What the heck 😶

gwynne: A somewhat bizarre mathematical quirk of the NIST curves that were chosen. See https://crypto.stackexchange.com/questions/6219/why-do-the-elliptic-curves-recommended-by-nist-use-521-bits-rather-than-512 for some gory details.

ptoffy: Actually that _is_ a typo, RSA is not based on elliptic curves. It should be 512 as that's the hashing algorithm

gwynne: Damn, you're right. Nice catch! (And doubly so with me incorrectly arguing the opposite 😅)

Tests/JWTKitTests/RSATests.swift
```diff
@@ -73,6 +73,57 @@ final class RSATests: XCTestCase {
...
+        XCTAssertEqual(exponent, publicExponent)
+    }
+
+    func testGetPrivateKeyPrimitivesWhenKeyIsPublicThrows() async throws {
+        let publicKey = try RSAKey(modulus: modulus, exponent: publicExponent)
+        XCTAssertThrowsError(try publicKey.getPrivateKeyPrimitives())
+    }
+
+    func testGetPrivateKeyPrimitivesFromNonRawBuiltKey() async throws {
+        let privateKey = try RSAKey.private(pem: """
```
gwynne, line 104: Can we also test that `BEGIN RSA PRIVATE KEY` will not decode as a PKCS#8 `PrivateKeyInfo` (or the public ones if that's easier) even if the data has the correct structure? (Conforming implementations are supposed to reject anything other than `BEGIN/END PRIVATE/PUBLIC KEY` for PKCS#8 PEM)

gwynne (commented):
Just one nit, looks pretty good! As for attribution to the test code copied from swift-asn1, update the `NOTICES.txt` file accordingly.

PR #129, opened 2023-11-28
Add customisable fields to JWTHeader

ptoffy:
Most credit goes to @gwynne

### DIFF
--- a/Sources/JWTKit/JWTHeader.swift
+++ b/Sources/JWTKit/JWTHeader.swift
@@ -9,9 +9,6 @@ public struct JWTHeader: Codable, Sendable {
     /// The Payload's Content Type.
     var cty: String?
 
-    /// Critical fields.
-    var crit: [String]?
-
     /// The JWT key identifier.
     package var kid: JWKIdentifier?
 
--- a/Sources/JWTKit/JWTHeader.swift
+++ b/Sources/JWTKit/JWTHeader.swift
@@ -1,5 +1,5 @@
 /// The header (details) used for signing and processing the JWT.
-package struct JWTHeader: Codable, Sendable {
+public struct JWTHeader: Codable, Sendable {
     /// The algorithm used with the signing.
     package var alg: String?
 
@@ -13,8 +13,56 @@ package struct JWTHeader: Codable, Sendable {
     var crit: [String]?
 
     /// The JWT key identifier.
-    var kid: JWKIdentifier?
+    package var kid: JWKIdentifier?
 
     /// The x5c certificate chain.
     package var x5c: [String]?
+
+    /// Custom fields.
+    package var customFields: [String: JWTHeaderField]?
+}
+
+public indirect enum JWTHeaderField: Hashable, Sendable, Codable {
+    case null
+    case bool(Bool)
+    case int(Int)
+    case string(String)
+    case array([JWTHeaderField])
+    case object([String: JWTHeaderField])
+
+    public init(from decoder: any Decoder) throws {
+        let container: any SingleValueDecodingContainer
+        do { container = try decoder.singleValueContainer() }
+        catch DecodingError.typeMismatch { self = .null; return }
+        if container.decodeNil() { self = .null; return }
+
+        do { self = try .bool(container.decode(Bool.self)); return }
+        catch DecodingError.typeMismatch {}
+
+        do { self = try .int(container.decode(Int.self)); return }
+        catch DecodingError.typeMismatch {}
+
+        do { self = try .string(container.decode(String.self)); return }
+        catch DecodingError.typeMismatch {}
+
+        do { self = try .array(container.decode([Self].self)); return }
+        catch DecodingError.typeMismatch {}
+
+        do { self = try .object(container.decode([String: Self].self)); return }
+        catch DecodingError.typeMismatch {}
+
+        throw DecodingError.dataCorruptedError(in: container, debugDescription: "No valid JSON type found.")
+    }
+
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.singleValueContainer()
+        switch self {
+        case .null: break
+        case let .bool(value): try container.encode(value)
+        case let .int(value): try container.encode(value)
+        case let .string(value): try container.encode(value)
+        case let .array(value): try container.encode(value)
+        case let .object(value): try container.encode(value)
+        }
+    }
 }
--- a/Sources/JWTKit/JWTSigner.swift
+++ b/Sources/JWTKit/JWTSigner.swift
@@ -22,9 +22,19 @@ final class JWTSigner: Sendable {
         typ: String = "JWT",
         kid: JWKIdentifier? = nil,
         cty: String? = nil,
-        x5c: [String]? = nil
+        x5c: [String]? = nil,
+        customFields: [String: JWTHeaderField]? = nil
     ) async throws -> String {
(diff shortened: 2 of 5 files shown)


Sources/JWTKit/JWTHeader.swift
```diff
@@ -9,12 +9,57 @@ package struct JWTHeader: Codable, Sendable {
     /// The Payload's Content Type.
     var cty: String?
 
-    /// Critical fields.
-    var crit: [String]?
-
     /// The JWT key identifier.
-    var kid: JWKIdentifier?
+    package var kid: JWKIdentifier?
```
gwynne, line 13: Package package package... 🤨

Tests/JWTKitTests/JWTKitTests.swift
gwynne, line 1: Your tests are whole inadequate - you're not verifying that the result of the encoding is a desirable header structure. Based on my reading of the code, if you actually decoded the token you got from these tests, it'd look like this:
```json
{
  "alg": "HS256",
  "typ": "JWT",
  "customFields": {
    "foo": "bar",
    "baz": 42
  }
}
```
When what you want is this:
```json
{
  "alg": "HS256",
  "typ": "JWT",
  "foo": "bar",
  "baz": 42
}
```
This is very much doable; you just need to provide Codable implementations for `JWTHeader` itself as well as `JWTHeaderField`. Hint: `String` is `CodingKeyRepresentable`, and you can use that and `KeyedDecodingContainerProtocol.allKeys` to make it really easy to decode all the fields the desired way, and all without having to hurt your brain figuring out how to use `superDecoder(forKey:)`.

gwynne (changes requested):

Sources/JWTKit/JWTHeader.swift
```diff
@@ -1,20 +1,126 @@
...
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.alg = try container.decodeIfPresent(String.self, forKey: .alg)
+        self.typ = try container.decodeIfPresent(String.self, forKey: .typ)
+        self.cty = try container.decodeIfPresent(String.self, forKey: .cty)
+        self.kid = try container.decodeIfPresent(JWKIdentifier.self, forKey: .kid)
+        self.x5c = try container.decodeIfPresent([String].self, forKey: .x5c)
+
+        let excludedKeys: Set<String> = ["alg", "typ", "cty", "kid", "x5c"]
```
gwynne, line 46: My only complaint is I don't like having these hardcoded here - can we make this a property on the `CodingKeys` enum instead?

Sources/JWTKit/JWTHeader.swift
```diff
@@ -1,20 +1,126 @@
...
     /// The JWT key identifier.
-    var kid: JWKIdentifier?
+    public var kid: JWKIdentifier?
 
     /// The x5c certificate chain.
-    package var x5c: [String]?
+    public var x5c: [String]?
+
+    /// Custom fields.
+    public var customFields: [String: JWTHeaderField]?
```
gwynne, line 19: There's no reason for this to be `Optional` - an empty dictionary is identical in meaning to a `nil` one and requires less code acrobatics.

Sources/JWTKit/JWTHeader.swift
```diff
@@ -1,20 +1,126 @@
...
+        case alg
+        case typ
+        case cty
+        case kid
+        case x5c
+        case custom(name: String)
+    }
+}
+
+public indirect enum JWTHeaderField: Hashable, Sendable, Codable {
```
gwynne, line 83: This should go in its own source file, I think

Sources/JWTKit/JWTSigner.swift
```diff
@@ -22,9 +22,19 @@ final class JWTSigner: Sendable {
         typ: String = "JWT",
         kid: JWKIdentifier? = nil,
         cty: String? = nil,
-        x5c: [String]? = nil
+        x5c: [String]? = nil,
+        customFields: [String: JWTHeaderField]? = nil
```
gwynne, line 26: Again, I'd prefer a non-optional with a  `[:]` default here

gwynne (changes requested):
Almost there!

Sources/JWTKit/JWTHeader.swift
```diff
@@ -57,13 +57,24 @@ extension JWTHeader: Codable {
...
 
         self.customFields = try container.allKeys
-            .filter { CodingKeys.allCases.contains($0) }
+            .filter { !CodingKeys.allCases.contains($0) }
             .reduce(into: [String: JWTHeaderField]()) { result, key in
                 result[key.stringValue] = try container.decode(JWTHeaderField.self, forKey: key)
             }
     }
 
     private enum CodingKeys: CodingKey, CaseIterable, Equatable {
```
gwynne, line 66: Don't conform to `CaseIterable` here - you're not actually fulfilling the protocol's requirements. You don't need to rename the property, just don't claim conformance to the protocol. (By declaring `CaseIterable` you're claiming that you're providing a complete and constant set of all possible enum cases, which isn't possible with a `String` as an associated value. You're also, more subtly, claiming that no two distinct cases are treated as equal, which is also not the case (for example, `.alg` and `.custom(name: "alg")` compare the same by raw value, but not by semantic - good thing you didn't let the compiler synthesize a `Hashable` conformance!). Violating the transitivity invariant is okay for `Equatable` in this context, but that's as far as I'd care to go.

Tests/JWTKitTests/InternalTests.swift
```diff
@@ -127,12 +130,9 @@ final class InternalTests: XCTestCase {
...
         }
         """
-        let jsonDecoder = JSONDecoder()
-        let encodedHeader = try JSONEncoder().encode(parser.header())
-        XCTAssertEqual(fields, customFields)
         XCTAssertEqual(
-            try jsonDecoder.decode([String: JWTHeaderField].self, from: encodedHeader),
-            try jsonDecoder.decode([String: JWTHeaderField].self, from: jsonFields.data(using: .utf8)!)
+            try JSONDecoder().decode([String: JWTHeaderField].self, from: encodedHeader),
+            try JSONDecoder().decode([String: JWTHeaderField].self, from: jsonFields.data(using: .utf8)!)
```
gwynne, line 135: This shouldn't change anything; if it does, something's wrong.

gwynne (changes requested):

Sources/JWTKit/JWTHeader.swift
gwynne: And while you're at it, you can change `allCases` to return a `Set<CodingKeys>` and then use `Set(container.allKeys).subtracting(CodingKeys.allCases)` above for an easy efficiency win.

Sources/JWTKit/JWTHeaderField.swift
```diff
@@ -0,0 +1,63 @@
+public indirect enum JWTHeaderField: Hashable, Sendable, Codable {
+    case null
```
0xTim, line 2: Are there definitely no other types we should support? This adds a new public enum 😛

0xTim (commented):
Some comments

Sources/JWTKit/JWTHeaderField.swift
gwynne: As it happens, you're right, I did (partially) miss one - JSON numbers aren't required to be integers; they can be decimal. JSON considers both integers and decimals to just be numbers, but the most convenient representation in Swift would treat them as separate "types" and add a `case float(Double)` case. (It's a hell of a lot easier than dealing with the 100% representative but bloody painful to handle `NSDecimal` way of doing it)

PR #130, opened 2023-12-06
Allow for custom JWT de/serialisation

ptoffy:


Sources/JWTKit/JWTHeaderField.swift
```diff
@@ -61,10 +44,98 @@ public indirect enum JWTHeaderField: Hashable, Sendable, Codable {
...
         }
     }
 }
+
+public extension JWTHeaderField {
+    internal var isNull: Bool { if case .null = self { true } else { false } }
+    var asBool: Bool? { if case let .bool(b) = self { b } else { nil } }
+    var asInt: Int? { if case let .int(i) = self { i } else { nil } }
+    var asDecimal: Double? { if case let .decimal(d) = self { d } else { nil } }
+    var asString: String? { if case let .string(s) = self { s } else { nil } }
```
0xTim, line 60: I'd prefer these be marked explicitly public just to make it clear

Sources/JWTKit/X5C/X5CVerifier.swift
```diff
@@ -118,16 +118,16 @@ public struct X5CVerifier: Sendable {
...
         // Ensure the algorithm used is ES256, as it's the only supported one (for now)
-        guard let headerAlg = header.alg, headerAlg == "ES256" else {
-            throw JWTError.invalidX5CChain(reason: "Unsupported algorithm: \(header.alg ?? "nil")")
+        guard let headerAlg = header.alg?.asString, headerAlg == "ES256" else {
+            throw JWTError.invalidX5CChain(reason: "Unsupported algorithm: \(String(describing: header.alg))")
         }
 
         // Ensure the x5c header parameter is present and not empty
-        guard let x5c = header.x5c, !x5c.isEmpty else {
+        guard let x5c = try header.x5c?.asArray(of: String.self), !x5c.isEmpty else {
```
0xTim, line 130: It might be nice to provide convenience extensions for stuff that's pretty common like this

Sources/JWTKit/JWTParser.swift
```diff
@@ -42,3 +34,27 @@ struct JWTParser: Sendable {
         encodedHeader + [.period] + encodedPayload
     }
 }
+
+public struct DefaultJWTParser: JWTParser {
```
0xTim, line 38: Given this is what will be used, I'm not a massive fan of the API, as it might get confusing. We could rename it to the type as `JWTParser` and make the protocol `CustomJWTParser` but I'm not sure that's better

Sources/JWTKit/JWTSerializer.swift
```diff
@@ -51,3 +30,34 @@ struct JWTSerializer {
         return String(decoding: bytes, as: UTF8.self)
     }
 }
+
+public struct DefaultJWTSerializer: JWTSerializer {
```
0xTim, line 34: As above with the parser comments

Tests/JWTKitTests/PSSTests.swift
```diff
@@ -20,7 +20,7 @@ final class PSSTests: XCTestCase {
             exp: .init(value: .init(timeIntervalSince1970: 2_000_000_000))
         )
 
-        let token = try await keyCollection.sign(payload, kid: "private")
+        let token = try await keyCollection.sign(payload, with: .init(fields: ["kid": .string("private")]))
```
0xTim, line 23: My gut tells me we should try and tighten up this API for the common use case, either via syntactic sugar or helper functions that retain the old API and ferry it down to the new API

0xTim (commented):
Overall I'm pretty happy with this actually. I'm not too concerned with how much it opens but I think we can iterate on the API to improve it for the common use cases a bit

0xTim (approved):

Package.swift
```diff
@@ -16,6 +16,8 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-crypto.git", from: "3.0.0"),
         .package(url: "https://github.com/apple/swift-certificates.git", branch: "main"),
         .package(url: "https://github.com/attaswift/BigInt.git", from: "5.3.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.0.0"),
```
0xTim, line 19: Do we need this dependency if we've copied everything in? It would be good not to have these as dependencies if possible

ptoffy: Ended up removing the test so we don't need that anymore

PR #135, opened 2024-02-14
Add RSA size boundary

ptoffy:
This adds the 2048 bit RSA key size lower boundary as of RFC 7518 section  [3.3](https://datatracker.ietf.org/doc/html/rfc7518#section-3.3) and [3.5](https://datatracker.ietf.org/doc/html/rfc7518#section-3.5)


0xTim:
@ptoffy this can be closed right as it's being upstreamed into Swift Crypto?

0xTim (approved):
Approving whilst we wait for the Crypto PR to be discussed

PR #143, opened 2024-03-14
add curve and parameters properties to ECDSAKey protocol

JaapWijnen:
This adds curve and parameter properties to the ECDSAKey protocol and also makes them public (non settable) on the ECDSA.PublicKey<Curve> and ECDSA.PrivateKey<Curve> types.


JaapWijnen:
Not sure what's going on with CI here. Any ideas? @ptoffy or @gwynne ?

ptoffy:
@JaapWijnen yeah looks like this isn't you, the same error is happening all over Vapor repos. There's probably something wrong with GH's CI or similar

ptoffy (changes requested):
While we're making these public, could you please move tests such as https://github.com/vapor/jwt-kit/blob/main/Tests/JWTKitTests/InternalTests.swift#L7 to the `ECDSATests` file so we can minify internal tests as much as possible? Thanks!

ptoffy (approved):

PR #145, opened 2024-03-15
make JWK.Curve Equatable again

JaapWijnen:
This makes sure that JWK.Curve is Equatable again


ptoffy (approved):
Good to merge once CI passes

PR #147, opened 2024-03-19
Add primary associated type to ECDSAKey

sidepelican:
Hello. I'm looking forward to jwt-kit v5 and trying it in my project.

I would like to propose adding a primary associated type (PAT) to `ECDSAKey` protocol. 
In my project, the key type is dynamic because it changes the type depending to environment variables.
Here is an example to illustrate the situation.

```swift
struct Key {
    func toECDSAKey() -> any ECDSAKey /* <P256> */  {
        // returns public or private key
    }
}

func setup(key: Key) async {
    let keys = JWTKeyCollection()
    await keys.addES256(key: key.toECDSAKey()) // Instance method 'addES256(key:kid:parser:serializer:)' requires the types 'Key.Curve' and 'P256' be equivalent
}
```

Currently, it is not possible to call `addES256` due to the lack of PAT in `ECDSAKey`. Incorporating PAT into `ECDSAKey` would resolve this issue.

### DIFF
--- a/Sources/JWTKit/ECDSA/ECDSA.swift
+++ b/Sources/JWTKit/ECDSA/ECDSA.swift
@@ -5,7 +5,7 @@ import X509
 
 public enum ECDSA: Sendable {}
 
-public protocol ECDSAKey: Sendable {
+public protocol ECDSAKey<Curve>: Sendable {
     associatedtype Curve: ECDSACurveType
     
     var curve: ECDSACurve { get }
--- a/Sources/JWTKit/ECDSA/JWTKeyCollection+ECDSA.swift
+++ b/Sources/JWTKit/ECDSA/JWTKeyCollection+ECDSA.swift
@@ -21,14 +21,12 @@ public extension JWTKeyCollection {
     ///          If `nil`, a default decoder is used.
     /// - Returns: The same instance of the collection (`Self`), which allows for method chaining.
     @discardableResult
-    func addES256<Key: ECDSAKey>(
+    func addES256<Key: ECDSAKey<P256>>(
         key: Key,
         kid: JWKIdentifier? = nil,
         parser: some JWTParser = DefaultJWTParser(),
         serializer: some JWTSerializer = DefaultJWTSerializer()
-    ) -> Self
-        where Key.Curve == P256
-    {
+    ) -> Self {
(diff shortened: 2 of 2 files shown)


ptoffy:
Hi, thanks for this, however as it adds complexity to the code and seems like a somewhat rare use case, I'm a bit reluctant in accepting the change. Have you tried something like
```swift
enum ES256Key {
    case `public`(ES256PublicKey)
    case `private`(ES256PrivateKey)
}
```

sidepelican:
Yes, I have temporarily implemented a solution similar to what you suggested.

However, at the heart of this issue is that `ECDSAKey` alone, unlike `RSAKey` or `EdDSAKey`, cannot be abstractly handled.

Looking at jwt-kit as a whole, the Signer abstractly handles both public and private keys. Yet, in the case of `ECDSAKey`, despite `ECDSASigner` abstractly managing the keys, the interface of `JWTKeyCollection` enforces a concrete handling.

This poses a problem when using `ECDSAKey`, as the usability of jwt-kit becomes unlike that of other keys. To align the usability with other keys, I would like to apply this change.

sidepelican:
@ptoffy ping

sidepelican:
Hello @ptoffy. Could you please share your thoughts?

I would like to understand why you think this change is complex. Are you aware that the primary associated type is not a generic protocol?

0xTim:
@sidepelican sorry this has been queued on me. We've just been discussing the implications of having another generic type, both whether it complicates stuff and whether we can actually use it to tidy up the API

We'll look into this properly today

ptoffy:
Hey @sidepelican, does this by any chance solve your problem? The update is probably going to get merged anyway so I thought I'd ask https://github.com/ptoffy/jwt-kit/blob/generic-add/Sources/JWTKit/ECDSA/JWTKeyCollection%2BECDSA.swift#L5
this allows for non-curve-specific keys to be added to a collection, the same method is available for the other algorithms too

sidepelican:
https://github.com/vapor/jwt-kit/pull/147#issuecomment-2049414968
@ptoffy Thank you for this. This approach does solve my problem. I'm fine with this direction as well.

ptoffy:
Closing this in favour of #160

PR #148, opened 2024-03-23
Make `JWTError.Backing` a `struct`

ptoffy:


gwynne (approved):
Nice change! Why was it ever like that to begin with? 🤦‍♀️

PR #150, opened 2024-03-25
Add initialisers to vendor identity tokens 

### LINKED ISSUE #131: Can't create `GoogleIdentityToken` without a Decoder
**Is your feature request related to a problem? Please describe.**
Often times we provide properties of `GoogleIdentityToken` manually, as we get them through Google Docs etc, not from a JSON body. But because it extends `Decodable`, the automatic initializer isn't available, so the only way to create it (AFAIK) is to create a custom `Decodable` object, which is awkward and bad in performance. We can't even extend it with our own initializer because it contains `let` properties.

**Describe the solution you'd like**
An initializer could be added manually, if no less-boilerplate solution is found.

**Describe alternatives you've considered**
`GoogleIdentityTokenFactory` object (with the same properties as `GoogleIdentityToken`) that doesn't conform to `Decodable` (thus has the automatic memberwise initializer) and has a method that returns `GoogleIdentityToken` (or `GoogleIdentityToken` could have an initializer taking `GoogleIdentityTokenFactory`, for discoverability). I think this is overengineering it though, and doesn't provide less boilerplate.

**Additional context**
I may be missing something, because this looks like a very commonly used thing.

ptoffy:
Fixes #131. Also adds tests to the vendor token payloads to get some coverage up

### DIFF
--- a/Sources/JWTKit/Claims/TenantIDClaim.swift
+++ b/Sources/JWTKit/Claims/TenantIDClaim.swift
@@ -8,7 +8,7 @@
 /// case-sensitive string representing a GUID. The presence of this claim and its proper validation
 /// are critical for the security of multi-tenant applications.
 public struct TenantIDClaim: JWTClaim, Equatable, ExpressibleByStringLiteral, ExpressibleByNilLiteral {
-    /// See ``JWTClaim``.
+    // See `JWTClaim.value`.
     public var value: String?
 
     // See `JWTClaim.init(value:)`.
--- a/Sources/JWTKit/Claims/TenantIDClaim.swift
+++ b/Sources/JWTKit/Claims/TenantIDClaim.swift
@@ -17,7 +17,7 @@ public struct TenantIDClaim: JWTClaim, Equatable, ExpressibleByStringLiteral, Ex
     }
 
     public init(stringLiteral value: String) {
-        self.value = value
+        self.init(value: value)
     }
 
     public init(nilLiteral: ()) {
--- a/Sources/JWTKit/Claims/TenantIDClaim.swift
+++ b/Sources/JWTKit/Claims/TenantIDClaim.swift
@@ -21,6 +21,6 @@ public struct TenantIDClaim: JWTClaim, Equatable, ExpressibleByStringLiteral, Ex
     }
 
     public init(nilLiteral: ()) {
-        self.value = nil
+        self.init(value: nil)
     }
 }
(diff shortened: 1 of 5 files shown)


Sources/JWTKit/Claims/TenantIDClaim.swift
```diff
@@ -7,12 +7,20 @@
...
     public init(value: String?) {
         self.value = value
     }
+
+    public init(stringLiteral value: String) {
+        self.value = value
+    }
+
+    public init(nilLiteral: ()) {
+        self.value = nil
```
gwynne, line 24: Delegate to the existing initializer.
```suggestion
        self.init(value: nil)
```

Sources/JWTKit/Claims/TenantIDClaim.swift
```diff
@@ -7,12 +7,20 @@
 /// logging or auditing the tenant context of the authenticated user. The value of "tid" is a
 /// case-sensitive string representing a GUID. The presence of this claim and its proper validation
 /// are critical for the security of multi-tenant applications.
-public struct TenantIDClaim: JWTClaim, Equatable {
+public struct TenantIDClaim: JWTClaim, Equatable, ExpressibleByStringLiteral, ExpressibleByNilLiteral {
     /// See ``JWTClaim``.
```
gwynne, line 11: When documenting properties or methods which implement protocol requirements, do _not_ use DocC comments; it suppresses autogeneration of the appropriate reference to the protocol.
```suggestion
    // See `JWTClaim.value`.
```

Sources/JWTKit/Claims/TenantIDClaim.swift
```diff
@@ -7,12 +7,20 @@
...
     public var value: String?
 
     /// See ``JWTClaim``.
```
gwynne, line 14: When documenting properties or methods which implement protocol requirements, do _not_ use DocC comments; it suppresses autogeneration of the appropriate reference to the protocol.
```suggestion
    // See `JWTClaim.init(value:)`.
```

Tests/JWTKitTests/VendorTokenTests.swift
gwynne, line 1: There is no test coverage added for `TenantIDClaim`.

ptoffy: I'm planning to add test coverage for claims in a separate PR

gwynne (approved):

PR #170, opened 2024-06-28
Allow iterating keys on key collection and adopt `swift-log`

ptoffy:
This addresses #169 and fixes a typo.
Also adds `swift-log` as a dependency to avoid `print`ing stuff

### DIFF
--- a/Sources/JWTKit/JWTKeyCollection.swift
+++ b/Sources/JWTKit/JWTKeyCollection.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Logging
 
 /// A collection of JWT and JWK signers for handling JSON Web Tokens (JWTs).
 ///
@@ -19,6 +20,8 @@ public actor JWTKeyCollection: Sendable {
     /// Whether the collection should iterate over all keys when verifying a JWT.
     let shouldIterateKeys: Bool
 
+    let logger: Logger
+
     /// Creates a new empty Signers collection.
     /// - parameters:
     ///    - shouldIterateKeys: Whether the collection should iterate over all keys when verifying a JWT.
@@ -27,12 +30,14 @@ public actor JWTKeyCollection: Sendable {
     public init(
         shouldIterateKeys: Bool = false,
         defaultJWTParser: some JWTParser = DefaultJWTParser(),
-        defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer()
+        defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer(),
+        logger: Logger = Logger(label: "jwt_kit_do_not_log", factory: { _ in SwiftLogNoOpLogHandler() })
     ) {
         self.storage = [:]
         self.shouldIterateKeys = shouldIterateKeys
         self.defaultJWTParser = defaultJWTParser
         self.defaultJWTSerializer = defaultJWTSerializer
+        self.logger = logger
     }
 
     /// Adds a ``JWTSigner`` to the collection, optionally associating it with a specific key identifier (KID).
@@ -49,12 +54,12 @@ public actor JWTKeyCollection: Sendable {
 
         if let kid {
             if self.storage[kid] != nil {
-                print("Warning: Overwriting existing JWT signer for key identifier '\(kid)'.")
+                logger.warning("Overwriting existing JWT signer", metadata: ["kid": "\(kid)"])
             }
             self.storage[kid] = .jwt(signer)
         } else {
             if self.default != nil {
-                print("Warning: Overwriting existing default JWT signer.")
+                logger.warning("Overwriting existing default JWT signer")
             }
             self.default = .jwt(signer)
         }
--- a/Sources/JWTKit/JWTKeyCollection.swift
+++ b/Sources/JWTKit/JWTKeyCollection.swift
@@ -16,12 +16,20 @@ public actor JWTKeyCollection: Sendable {
     public let defaultJWTParser: any JWTParser
     public let defaultJWTSerializer: any JWTSerializer
 
+    /// Whether the collection should iterate over all signers when verifying a JWT.
+    let shouldIterateKeys: Bool
+
     /// Creates a new empty Signers collection.
     /// - parameters:
     ///    - jsonEncoder: The default JSON encoder.
     ///    - jsonDecoder: The default JSON decoder.
-    public init(defaultJWTParser: some JWTParser = DefaultJWTParser(), defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer()) {
+    public init(
+        shouldIterateKeys: Bool = false,
+        defaultJWTParser: some JWTParser = DefaultJWTParser(),
+        defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer()
+    ) {
         self.storage = [:]
+        self.shouldIterateKeys = shouldIterateKeys
         self.defaultJWTParser = defaultJWTParser
         self.defaultJWTSerializer = defaultJWTSerializer
     }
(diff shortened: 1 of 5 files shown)


Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -16,12 +16,20 @@ public actor JWTKeyCollection: Sendable {
     public let defaultJWTParser: any JWTParser
     public let defaultJWTSerializer: any JWTSerializer
 
+    /// Whether the collection should iterate over all signers when verifying a JWT.
+    let shouldIterateKeys: Bool
+
     /// Creates a new empty Signers collection.
     /// - parameters:
     ///    - jsonEncoder: The default JSON encoder.
     ///    - jsonDecoder: The default JSON decoder.
```
gwynne, line 25: Add documentation for the new parameter here

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -16,12 +16,20 @@ public actor JWTKeyCollection: Sendable {
     public let defaultJWTParser: any JWTParser
     public let defaultJWTSerializer: any JWTSerializer
 
+    /// Whether the collection should iterate over all signers when verifying a JWT.
+    let shouldIterateKeys: Bool
```
gwynne, line 20: Which is it, all signers or all keys?

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -205,8 +213,26 @@ public actor JWTKeyCollection: Sendable {
...
+                    do {
+                        signer = try self.getSigner(for: kid, alg: header.alg)
+                        return try await signer.verify(token)
+                    } catch {
+                        continue
+                    }
+                }
+
+                throw error
+            }
```
gwynne, line 234:
```suggestion
            if self.shouldIterateKeys {
                for (kid, _) in self.storage {
                    do {
                        signer = try self.getSigner(for: kid, alg: header.alg)
                        return try await signer.verify(token)
                    } catch {}
                }
            }
            throw error
```

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -38,7 +46,7 @@ public actor JWTKeyCollection: Sendable {
     func add(_ signer: JWTSigner, for kid: JWKIdentifier? = nil) -> Self {
         let signer = JWTSigner(algorithm: signer.algorithm, parser: signer.parser, serializer: signer.serializer)
 
-        if let kid = kid {
+        if let kid {
             if self.storage[kid] != nil {
                 print("Warning: Overwriting existing JWT signer for key identifier '\(kid)'.")
```
gwynne, line 51: Is there a `Logger` that can be used here instead of `print()`?

gwynne (changes requested):

Sources/JWTKit/JWTKeyCollection.swift
ptoffy: Not really. Adding swift-log for that might not really be worth it

gwynne: Agreed, but at the same time IIRC it's considered bad form for libraries to `print()`. @0xTim Do you know what the recommendation is in such a case?

gwynne (approved):

Sources/JWTKit/JWTKeyCollection.swift
0xTim: Yeah printing is a no go, we need to use a logger if we want to log stuff. I'm surprised we haven't needed to up to this point

0xTim (changes requested):
Use a logger instead of printing

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -205,8 +219,21 @@ public actor JWTKeyCollection: Sendable {
...
-        return try await signer.verify(token)
+        var signer = try self.getSigner(for: kid, alg: header.alg)
+
+        do {
+            return try await signer.verify(token)
+        } catch {
+            if self.shouldIterateKeys {
+                for (kid, _) in self.storage {
+                    do {
+                        signer = try self.getSigner(for: kid, alg: header.alg)
```
0xTim, line 230: What happens if a signer doesn't have a `kid`?

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -205,8 +219,21 @@ public actor JWTKeyCollection: Sendable {
...
         let header = try defaultJWTParser.parseHeader(token)
         let kid = header.kid.flatMap { JWKIdentifier(string: $0) }
-        let signer = try self.getSigner(for: kid, alg: header.alg)
-        return try await signer.verify(token)
+        var signer = try self.getSigner(for: kid, alg: header.alg)
+
+        do {
+            return try await signer.verify(token)
+        } catch {
+            if self.shouldIterateKeys {
```
0xTim, line 227: Minor nit we're verifying the signer twice here

0xTim (changes requested):
Do we have any other prints?

Sources/JWTKit/JWTKeyCollection.swift
ptoffy: It's marked as default. Adding a different signer without a kid will override the one before

ptoffy:
@0xTim no there's no other prints

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -38,14 +47,14 @@ public actor JWTKeyCollection: Sendable {
     func add(_ signer: JWTSigner, for kid: JWKIdentifier? = nil) -> Self {
         let signer = JWTSigner(algorithm: signer.algorithm, parser: signer.parser, serializer: signer.serializer)
 
-        if let kid = kid {
+        if let kid {
             if self.storage[kid] != nil {
-                print("Warning: Overwriting existing JWT signer for key identifier '\(kid)'.")
+                logger.warning("Overwriting existing JWT signer", metadata: ["kid": "\(kid)"])
```
0xTim, line 52: These should be debug as per the log level guide https://www.swift.org/documentation/server/guides/libraries/log-levels.html

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -38,14 +47,14 @@ public actor JWTKeyCollection: Sendable {
...
             }
             self.storage[kid] = .jwt(signer)
         } else {
             if self.default != nil {
-                print("Warning: Overwriting existing default JWT signer.")
+                logger.warning("Overwriting existing default JWT signer")
```
0xTim, line 57: As above

0xTim (changes requested):

Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -16,14 +17,21 @@ public actor JWTKeyCollection: Sendable {
...
+
     /// Creates a new empty Signers collection.
     /// - parameters:
     ///    - jsonEncoder: The default JSON encoder.
     ///    - jsonDecoder: The default JSON decoder.
-    public init(defaultJWTParser: some JWTParser = DefaultJWTParser(), defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer()) {
+    public init(
+        defaultJWTParser: some JWTParser = DefaultJWTParser(),
+        defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer(),
+        logger: Logger = Logger(label: "jwt_kit_do_not_log", factory: { _ in SwiftLogNoOpLogHandler() })
```
0xTim, line 29: Sorry I should have added we should have this documented as well

0xTim (approved):

PR #171, opened 2024-07-02
update example code and convert to Snippets

tayloraswift:
i converted about half of the example code to SwiftPM Snippets, and updated the names of the functions so that the examples would compile once more. 

i previewed it in DocC, but on Swiftinit, it would also render with linked symbols in the code.

the final examples on the page do not compile due to calling internal API, so i left them out for this PR.

```
/swift/jwt-kit/Snippets/Customization.swift:9:50: error: 'base64URLEncodedBytes' is inaccessible due to 'package' protection level
            try Data(jsonEncoder.encode(payload).base64URLEncodedBytes())
                                                 ^~~~~~~~~~~~~~~~~~~~~
/swift/jwt-kit/Sources/JWTKit/Utilities/Base64URL.swift:19:10: note: 'base64URLEncodedBytes()' declared here
    func base64URLEncodedBytes() -> [UInt8] {
         ^
/swift/jwt-kit/Snippets/Customization.swift:22:87: error: 'base64URLDecodedBytes' is inaccessible due to 'package' protection level
        let header = try jsonDecoder.decode(JWTHeader.self, from: .init(encodedHeader.base64URLDecodedBytes()))
                                                                                      ^~~~~~~~~~~~~~~~~~~~~
/swift/jwt-kit/Sources/JWTKit/Utilities/Base64URL.swift:15:10: note: 'base64URLDecodedBytes()' declared here
    func base64URLDecodedBytes() -> [UInt8] {
         ^
/swift/jwt-kit/Snippets/Customization.swift:30:47: error: 'base64URLDecodedBytes' is inaccessible due to 'package' protection level
        let signature = Data(encodedSignature.base64URLDecodedBytes())
                                              ^~~~~~~~~~~~~~~~~~~~~
/swift/jwt-kit/Sources/JWTKit/Utilities/Base64URL.swift:15:10: note: 'base64URLDecodedBytes()' declared here
    func base64URLDecodedBytes() -> [UInt8] {
```

### DIFF
--- a/Snippets/JWKExamples.swift
+++ b/Snippets/JWKExamples.swift
@@ -1,0 +1,16 @@
+import Foundation
+import JWTKit
+
+let rsaModulus = "..."
+
+let json = """
+{
+    "keys": [
+        {"kty": "RSA", "alg": "RS256", "kid": "a", "n": "\(rsaModulus)", "e": "AQAB"},
+        {"kty": "RSA", "alg": "RS512", "kid": "b", "n": "\(rsaModulus)", "e": "AQAB"},
+    ]
+}
+"""
+
+// Create key collection and add JWKS
+let keys = try await JWTKeyCollection().use(jwksJSON: json)
--- a/Snippets/JWTKitExamples.swift
+++ b/Snippets/JWTKitExamples.swift
@@ -23,6 +23,15 @@ await keys.add(hmac: "secret", digestAlgorithm: .sha256)
 // Registers an HS256 (HMAC-SHA-256) signer with a key identifier.
 await keys.add(hmac: "secret", digestAlgorithm: .sha256, kid: "my-key")
 
+// snippet.KEY_COLLECTION_CREATE_ES256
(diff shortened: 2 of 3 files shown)


ptoffy:
@0xTim I feel like the header customisation stuff is quite useful/interesting to have as an example there. The internal methods are internal because I feel like it shouldn't be JWTKit's place to provide such helper methods, but at the same time they are required for that example (and probably for most use cases of that API) to run so adding them in isn't such a bad idea IMO

tayloraswift:
i outlined the remaining code from `index.md` following the advice of this thread, and generated a temporary preview on Swiftinit you can review here:

https://swiftinit.org/hist/tayloraswift.jwt-kit:snippets/jwtkit#Getting%20Started

although i didn’t want to introduce a new dependency in this PR, i’ll also note for posterity that swift-hash provides a pure-Swift implementation of URL-friendly Base64 encoding/decoding here:

https://swiftinit.org/docs/swift-hash/base64/base64.encode(_:padding:with:)

ptoffy (approved):
Thanks!

PR #172, opened 2024-07-22
Cache JWKSigner when possible

ptoffy:
Instead of creating a new signer for JWKs each time, try caching it if we know the algorithm upfront


Sources/JWTKit/JWTKeyCollection.swift
```diff
@@ -29,7 +29,7 @@ public actor JWTKeyCollection: Sendable {
         defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer(),
         logger: Logger = Logger(label: "jwt_kit_do_not_log", factory: { _ in SwiftLogNoOpLogHandler() })
     ) {
-        self.storage = [:]
+        storage = [:]
```
0xTim, line 32: Why change this? It's inconsistent with the rest of the lines

0xTim (approved):

Sources/JWTKit/JWTKeyCollection.swift
ptoffy: Damn formatter

PR #177, opened 2024-07-23
Move to SwiftCrypto EdDSA key

### LINKED ISSUE #174: JWTKit 5.0.0-RC.1 support for Curve25519 types
JWTKit's RC uses its own types to represent common cryptographic types, and then convert into Swift-Crypto types. It would be much more practical, especially when used with other libraries, to use the swift-crypto types themselves. This way we can plug in the private keys we already have, without going through a base64 conversion step and back.

ptoffy:
Fix #174

### DIFF
--- a/Snippets/JWTKitExamples.swift
+++ b/Snippets/JWTKitExamples.swift
@@ -10,7 +10,7 @@ struct ExamplePayload: JWTPayload {
     var exp: ExpirationClaim
     var admin: BoolClaim
 
-    func verify(using key: some JWTAlgorithm) throws {
+    func verify(using _: some JWTAlgorithm) throws {
         try self.exp.verifyNotExpired()
     }
 }
@@ -33,8 +33,7 @@ let key = try ES256PublicKey(pem: ecdsaPublicKey)
 await keys.add(ecdsa: key)
 
 // snippet.end
-do
-{
+do {
     // Create a new instance of our JWTPayload
     let payload = ExamplePayload(
         sub: "vapor",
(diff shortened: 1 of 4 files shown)


Snippets/JWTKitExamples.swift
```diff
@@ -63,14 +61,13 @@ do
...
 
-do
-{
+do {
     // snippet.EDDSA
     // Initialize an EdDSA key with public PEM
     let publicKey = try EdDSA.PublicKey(x: "...", curve: .ed25519)
 
     // Initialize an EdDSA key with private PEM
-    let privateKey = try EdDSA.PrivateKey(x: "...", d: "...", curve: .ed25519)
```
0xTim, line 73: Do we not need this in most cases?

Sources/JWTKit/EdDSA/EdDSA.swift
```diff
@@ -13,23 +13,19 @@ public extension EdDSA {
     /// A struct representing a public key used in EdDSA (Edwards-curve Digital Signature Algorithm).
     ///
     /// In JWT, EdDSA public keys are represented as a single x-coordinate and are used for verifying signatures.
-    /// Currently, only the ``Curve/ed25519`` curve is supported.
+    /// Currently, only the ``EdDSACurve/ed25519`` curve is supported.
     struct PublicKey: EdDSAKey {
-        // https://www.rfc-editor.org/rfc/rfc8037.html#appendix-A.2
```
0xTim, line 18: Out of interest, should this be public?

0xTim (approved):
Couple of comments otherwise LGTM

Snippets/JWTKitExamples.swift
ptoffy: The init is still there, it just was renamed to `init(d:, curve:)` because the `x` represents the public key and isn't needed there

Sources/JWTKit/EdDSA/EdDSA.swift
ptoffy: The struct is in a public extension so it is effectively public if not specified otherwise (otherwise the tests wouldn't pass)

0xTim: I do not enjoy this implicit access control in Swift 🙃

PR #181, opened 2024-07-29
Add compatibility for iOS 15 and friends

### LINKED ISSUE #180: Lower minimum iOS version
We'd love to use JWTKit for some of our iOS code, but we need to support iOS 15. Is it possible and feasible to lower the iOS requirement? A quick check doesn't seem to indicate any iOS 16+ dependencies.

ptoffy:
Fix #180

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,9 +5,9 @@ let package = Package(
     name: "jwt-kit",
     platforms: [
         .macOS(.v13),
-        .iOS(.v16),
-        .tvOS(.v16),
-        .watchOS(.v9),
+        .iOS(.v15),
+        .tvOS(.v15),
+        .watchOS(.v8),
     ],
     products: [
         .library(name: "JWTKit", targets: ["JWTKit"]),
--- a/Package@swift-5.10.swift
+++ b/Package@swift-5.10.swift
@@ -5,9 +5,9 @@ let package = Package(
     name: "jwt-kit",
     platforms: [
         .macOS(.v13),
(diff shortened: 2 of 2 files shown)


Package.swift
```diff
@@ -5,9 +5,9 @@ let package = Package(
     name: "jwt-kit",
     platforms: [
         .macOS(.v13),
```
0xTim, line 7: Do we want to bump this to 12 to line up?

Joannis (commented):
Thanks! Looks good to me. I'd recommend lowering this as much as feasible, going even back to iOS 13 if possible. You can always use `@available` annotations if need be for specific features.

Package.swift
ptoffy: There's a few errors around EC keys about missing conformances with `.macOS` < 13. Also `JSONDecoder` and `JSONEncoder` are only `Sendable` in 13+

ptoffy: I mean we can try solving them but not sure it's worth it?

0xTim: What are the issues with EC keys? Because surely we'll have the same issues on iOS 15

0xTim: I wonder if it's worth being _really_ annoying and adding CI for iOS (including 15) so see what issues we'd hit

ptoffy: The thing is Crypto doesn't specify iOS requirements (or any besides macOS) so there's no issues with that

0xTim (approved):

PR #185, opened 2024-08-15
Last of the Swift 6 language mode warnings

0xTim:
Some more retroactive conformances


Package.swift
```diff
@@ -40,5 +40,5 @@ let package = Package(
             ]
         ),
     ],
-    swiftLanguageVersions: [.v6]
+    swiftLanguageModes: [.v6]
```
MahdiBM, line 43: Is this necessary? what does it do to explicitly declare it? The Swift version up there is already 6.

0xTim: The change is because it's been renamed in the latest betas.

We need to set the language mode because it's separate from the tools version and opt-in given it enforces strict concurrency checking

ptoffy (approved):

Package.swift
MahdiBM: Alright I knew about the rename but I thought having a 6.0 Package.swift should be enough. Still need to check for myself.

PR #189, opened 2024-09-01
Fix decimal and nil header encoding

ptoffy:
This aims to fix the issues discovered in #188
1. `nil` wasn't being added because assigning `nil` to a field would remove it from the dict. Now assigning `nil` to a field actually adds an explicit null value to the json, while there's a `remove(field:)` method to remove fields from the header
2. `decimal` wasn't being decoded correctly because it was trying to unwrap `Double`s as `Int`s


fpseverino (approved):
LGTM

0xTim (approved):

PR #192, opened 2024-09-11
Add availability annotations to unavailable crypto properties

### LINKED ISSUE #191: v5 not building for iOS project
### Describe the issue

Hitting compile issues when trying to build an iOS project with the v5 branch.

### Vapor version

5.0.0-rc.2

iOS 18.0

### Swift version

Swift Package Manager - Swift 6.0.0-dev

### Steps to reproduce

Getting various issues around `compressedRepresentation`. The iOS project I'm building is targeting iOS 16 and above.

```Protocol 'ECDSAPublicKey' requires 'compressedRepresentation' to be available in iOS 15.0 and newer```

### Outcome

It should build.

### Additional notes

Not sure if it's the same issue but it's also failing for build for iOS over at SPI https://swiftpackageindex.com/builds/C2B4497A-C8DE-40AD-A273-A504A3D3F31D

ptoffy:
Since adding support for iOS 15, builds started failing at https://swiftpackageindex.com/builds/495EA704-A76E-4078-BAA9-617D0769824E because of an unavailable property (namely `compressedRepresentation`) which is available from iOS 16 on. This adds annotations to make the property unavailable to iOS 15 (and therefore using Crypto's implementation)

Fixes #191

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,5 +36,7 @@ jobs:
     if: ${{ !(github.event.pull_request.draft || false) }}
     runs-on: macos-latest
     steps:
-      - name: Run iOS Tests on iOS
+      - name: Check out JWTKit
+        uses: actions/checkout@v4
+      - name: Run iOS Tests
         run: xcodebuild -scheme jwt-kit -destination generic/platform=iOS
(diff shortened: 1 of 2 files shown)


0xTim (changes requested):
@ptoffy looks like my iOS tests aren't working correctly - this should be failing on main. I think I picked an OS that isn't supported judging by the build log.

Could you remove the fixes, get CI to fail and then re-add them?

ptoffy:
Cool @0xTim we should be good

0xTim (approved):

PR #193, opened 2024-09-11
Show warnings as errors

ptoffy:


0xTim (approved):
Looks good but looks like we have some issues to fix 😆

PR #199, opened 2024-10-04
Support Crypto 4.0.0

0xTim:
Add support for the upcoming release of Swift Crypto 4.0.0


gwynne (approved):

0xTim:
There's a debate to be had over whether we just jump straight to 4.0.0 and drop support for 3.0.0 but I don't see an issue with supporting both.

(The nice thing is that we can decide at a later date since there are no more exports 🥳)

ptoffy (approved):

PR #212, opened 2024-11-03
Update certs for new ones

0xTim:
Update the certs and tokens using the scripts from `main` as the current ones are expired


ptoffy (approved):

0xTim:
API breakage is expected because it's testing against main.

Integration tests are failing on latest v4 tag and main due to an expired SIWA token I think. Will merge to get a release out

PR #217, opened 2024-11-11
Fix crash when verifying corrupted token

### LINKED ISSUE #213: Crash when verifying a corrupt token
### Describe the issue

Attempting to verify a corrupt token results in the process crashing.

### Vapor version

(unrelated to Vapor)

macOS 14.7.1

### Swift version

Swift Package Manager - Swift 6.0.3-dev

### Steps to reproduce

1. Generate a valid token
2. Randomly change a character in the generated token header, making the base64 string invalid
3. Attempt to `verify` the corrupt token using `verify(token)`

Below is a simple cli implementation that reproduces the issue:

> EDIT: See [comment](https://github.com/vapor/jwt-kit/issues/213#issuecomment-2465906938) for a smaller example that can reproduce the crash more reliably.

```swift
import Foundation
import JWTKit

struct TestPayload: JWTPayload {
    var sub: SubjectClaim
    var exp: ExpirationClaim
    var flag: Bool

    func verify(using algorithm: some JWTAlgorithm) async throws {
        try exp.verifyNotExpired()
    }
}

@main
struct JWTKitBugTest {
    static func main() async throws {
        do {
            let keys = JWTKeyCollection()
            await keys.add(hmac: "crashme", digestAlgorithm: .sha256)

            let test = TestPayload(sub: "hello", exp: .init(value: Date.now.addingTimeInterval(100)), flag: true)

            let token = try await keys.sign(test)

            print("Generated token: \(token)")

            _ = try await keys.verify(token, as: TestPayload.self)

            print("Verified unmodified token")

            var corruptToken = token

            /// Change a character in the token to `x`, invalidating the base64 string.
            corruptToken.replaceSubrange(token.index(token.startIndex, offsetBy: 24)..<token.index(token.startIndex, offsetBy: 25), with: "x")

            print("Corrupt token: \(corruptToken)")

            /// This crashes:
            _ = try await keys.verify(corruptToken, as: TestPayload.self, iteratingKeys: true)
        } catch {
            print(error)
        }
    }
}
```

### Outcome

The process will crash with the fatal error below, even if the caller did not use `try!` and was calling within a `do/catch` block:

```Thread 2: Fatal error: 'try!' expression unexpectedly raised an error: Foundation.JSONError.cannotConvertInputStringDataToUTF8(location: Foundation.JSONError.SourceLocation(line: 1, column: 17, index: 16))```

### Additional notes

The crash will not occur every time with the example above, but can be reliably reproduced by repeating the same crashing token after you've found one that does crash.

### PR DESCRIPTION
**Pull request by ptoffy**
Fix #213

### CODE REVIEW

File: Sources/JWTKit/JWTParser.swift [hunk 1]
```diff
@@ -58,9 +59,20 @@ public struct DefaultJWTParser: JWTParser {
         let payload: Payload
         let signature: Data
 
+        func isUTF8(_ bytes: [UInt8]) -> Bool {
+            bytes.allSatisfy { $0 < 128 }
+        }
```
gwynne, line 64: This is inaccurately named - this verifies `isASCII`, not `isUTF8`.

Sources/JWTKit/JWTParser.swift
```diff
@@ -58,9 +59,20 @@ public struct DefaultJWTParser: JWTParser {
         let payload: Payload
         let signature: Data
 
+        func isUTF8(_ bytes: [UInt8]) -> Bool {
+            String(bytes: bytes, encoding: .utf8) != nil
```
gwynne, line 63:
```suggestion
            String(validating: bytes, as: UTF8.self) != nil
```

gwynne (changes requested):

gwynne (approved):

PR #218, opened 2024-11-18
Add benchmarks

ptoffy:


0xTim (approved):

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,43 @@
...
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+      - name: Run benchmarks for PR branch
+        continue-on-error: true
+        run: |
+          swift package -c release --package-path Benchmarks --disable-sandbox benchmark baseline update pull_request
+      - name: Run benchmarks for 'main' branch
+        run: |
+          git stash
+          git checkout main
+          swift package -c release --package-path Benchmarks --disable-sandbox benchmark baseline update main
```
MahdiBM, line 30: We should use static thresholds like SwiftNIO does.
Joakim recently introduced a new command for that as well.

Benchmarks/Benchmarks/Signing/Signing.swift
```diff
@@ -0,0 +1,106 @@
+import Benchmark
+import Foundation
+import JWTKit
+
+let customThresholds = BenchmarkThresholds(
```
MahdiBM, line 5: I'd prefer this to be renamed to `thresholdTolerances`.
I talked with Joakim and he did confirm this is kind of a misnomer that the parameter is plainly named `threhsolds`.

.github/workflows/benchmark.yml
MahdiBM: Though static thresholds are only supported for p90 i think?
That's also what SwiftNIO is fine with.

Benchmarks/Benchmarks/TokenLifecycle/TokenLifecycle.swift
```diff
@@ -0,0 +1,139 @@
+import Benchmark
+import Foundation
+import JWTKit
+
+let customThresholds = BenchmarkThresholds(
+    relative: [.p25: 15.0, .p50: 15.0, .p75: 15.0, .p90: 15.0, .p99: 15.0],
+    absolute: [:]
```
MahdiBM, line 7: I'm not 100% sure but i think declaring `relative`, you still need to declare `absolute` as well?
Unless you've tried this before and know that it'll work.

MahdiBM (commented):
Some comments. I'm fine with merging this as-is and creating follow-up PRs.

Benchmarks/Benchmarks/TokenLifecycle/TokenLifecycle.swift
ptoffy: Actually IIRC this was done like this because we were running them on GH runners which were flaky and didn't work. It probably makes sense to use the defaults now

PR #220, opened 2024-11-19
Update benchmark CI to run only on approved PRs and Swift changes

ptoffy:


gwynne (approved):

.github/workflows/benchmark.yml
```diff
@@ -1,11 +1,17 @@
...
 on:
   workflow_dispatch:
+  pull_request_review:
+    types: [submitted]
   pull_request:
     branches: [ main ]
-  
+    types: [synchronize]
+    paths:
+      - '**/*.swift'
```
MahdiBM, line 10: add the CI file itself as well?

ptoffy: Mhh why?

MahdiBM: If we change the CI file we'd want to make sure it is still functional?

0xTim: Yes this. We can also ignore the tests as well

0xTim (approved):
As Mahdi says, we should update this to check for the benchmark files as well. But otherwise LGTM

PR #221, opened 2024-11-20
Version 4 to Version 5 Migration Deprecations

dimitribouniol:
This PR adds migration deprecations for many common APIs between version 4 and version 5 that should help most users (particularly those using ECDSA keys and standard JWTs) migrate from version 4 to version 5 by following warnings. Crucially, it allows libraries that rely on the same requirements (for instance, https://github.com/apple/app-store-server-library-swift and https://github.com/m-barthelemy/AcmeSwift) to allow for a range of versions such as `"4.14.0" ..< "6.0.0"`.

If we agree with the proposed updates here, a few more things should be done:
- Cut a new _minor_ v4 version, specifically `4.14.0`, so users can opt out of the warnings should they wish to.
- Add an entry to main's README explaining the process to update from v4 to v5.
- Update v5 with an `AsyncJWTPayload` typealias back to `JWTPayload` that is itself deprecated, to fully support libraries that wish to be compatible with v4 and v5.

Note that I didn't make compatibility APIs for _every_ type — if someone feels they are necessary due to a library requiring them, I figured we could add more support in a future PR.


Sources/JWTKit/JWTSigners.swift
```diff
@@ -280,3 +295,231 @@ private struct JWKSigner {
         }
     }
 }
+
+/// JWTKeyCollection was introduced in v5 and replaces ``JWTSigners``.
+///
+/// - Note: Please migrate over to ``JWTKeyCollection`` before updating to v5, though if you plan on remaining on v4, ``JWTSigners`` can continue to be used.
+public actor JWTKeyCollection {
+    var signers = JWTSigners()
```
dimitribouniol, line 303: This is a lightweight implementation of the new JWTKeyCollection API falling back entirely on the now deprecated JWTSigners API.

Sources/JWTKit/JWTPayload.swift
```diff
@@ -1,6 +1,17 @@
...
+
+/// A transitionary protocol with sync and async requirements.
+///
+/// This protocol should be dropped once you are finished migrating to v5, as it'll have been renamed back to ``JWTPayload``, but with a single async requirement. In order to support both versions v4 and v5 in a library, do not implement the requirements of ``JWTPayload`` as ``JWTSigner`` is no longer available in v5.
+public protocol AsyncJWTPayload: Codable {
+    func verify<Algorithm: JWTAlgorithm>(using signer: Algorithm) throws
+    
+    /// Verifies that the payload's claims are correct or throws an error.
+    func verify<Algorithm: JWTAlgorithm>(using algorithm: Algorithm) async throws
+}
```
dimitribouniol, line 17: Because we can't mark protocol methods as deprecated, we define a new one that will be transitionary and support both sync and asynchronous modalities, ensuring compatibility with existing code paths, but also newer signing methods no matter which would be used.

Sources/JWTKit/JWTSigners.swift
```diff
@@ -280,3 +295,231 @@ private struct JWKSigner {
...
+    ///   - header: An optional header to include in the JWT.
+    /// - Throws: An error if the payload cannot be signed.
+    /// - Returns: A signed JWT token string.
+    @available(*, deprecated, message: "Please make sure Payload conforms to AsyncJWTPayload instead of JWTPayload before updating to v5.")
+    public func sign<Payload: JWTPayload>(
+        _ payload: Payload,
+        kid: JWKIdentifier? = nil
+    ) async throws -> String {
+        try signers.sign(payload, kid: kid)
+    }
```
dimitribouniol, line 506: This variant is marked as deprecated despite appearing on the new JWTKeyCollection to help push users to adopt AsyncJWTPayload wherever they use another payload type.

Sources/JWTKit/JWTPayload.swift
```diff
@@ -1,6 +1,17 @@
...
+/// - Warning: Requirements changed in v5 to be async. Please also conform to ``AsyncJWTPayload`` while on v4, and remove the ``AsyncJWTPayload`` conformance once you do update to v5.
 public protocol JWTPayload: Codable {
     /// Verifies that the payload's claims are correct or throws an error.
     func verify(using signer: JWTSigner) throws
 }
+
+/// A transitionary protocol with sync and async requirements.
+///
+/// This protocol should be dropped once you are finished migrating to v5, as it'll have been renamed back to ``JWTPayload``, but with a single async requirement. In order to support both versions v4 and v5 in a library, do not implement the requirements of ``JWTPayload`` as ``JWTSigner`` is no longer available in v5.
+public protocol AsyncJWTPayload: Codable {
```
dimitribouniol, line 12: As for the name, I'm open to alternatives. `JWTPayloadV4` could work well since this variant of it only really exists in the v4 version of the codebase.

Sources/JWTKit/JWTParser.swift
```diff
@@ -30,6 +30,13 @@ struct JWTParser {
             .decode(Payload.self, from: .init(self.encodedPayload.base64URLDecodedBytes()))
     }
 
+    func payload<Payload>(as payload: Payload.Type, jsonDecoder: any JWTJSONDecoder) throws -> Payload
+        where Payload: AsyncJWTPayload
+    {
+        try jsonDecoder
+            .decode(Payload.self, from: .init(self.encodedPayload.base64URLDecodedBytes()))
+    }
```
dimitribouniol, line 38: This is a duplicate of the above variant to support the new protocol.

Sources/JWTKit/JWTSerializer.swift
```diff
@@ -9,7 +9,7 @@ struct JWTSerializer {
         cty: String? = nil,
         jsonEncoder: any JWTJSONEncoder
     ) throws -> String
-        where Payload: JWTPayload
+        where Payload: Encodable
```
dimitribouniol, line 12: Similarly, rather than duplicate this internal method, I loosened the protocol requirements to it works for both.

Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -190,3 +199,197 @@ extension ECDSAKey.Curve {
...
+public typealias ES512PublicKey = ECDSA.PublicKey<P521>
+public typealias ES512PrivateKey = ECDSA.PrivateKey<P521>
+
+public enum ECDSA: Sendable {
+    /// ECDSA.PublicKey was introduced in v5 and replaces ``ECDSAKey``.
+    ///
+    /// - Note: Please migrate over to ``ECDSA/PublicKey`` before updating to v5, though if you plan on remaining on v4, ``ECDSAKey`` can continue to be used.
+    public struct PublicKey<Curve: ECDSACurveType> {
+        let key: ECDSAKey
+        init(key: ECDSAKey) { self.key = key }
```
dimitribouniol, line 231: These are wrappers on the existing keys to support the new types a user would be forced into, wile not adding much new functionality. I added bare minimum support for all ECDSA public and private keys.

Tests/JWTKitTests/JWTMigrationTests.swift
```diff
@@ -0,0 +1,53 @@
+import XCTest
+import Crypto
+#if os(Linux)
+import FoundationNetworking
+#endif
+@testable import JWTKit
+
+class JWTKitMigrationTests: XCTestCase {
+    func testVerifyingCryptoKey() async throws {
```
dimitribouniol, line 9: An end to end test to validate everything is hooked up properly. Also tested the migrations in the app-store-server-library repo along with my own codebase.

Sources/JWTKit/ECDSA/ECDSAKey.swift
```diff
@@ -190,3 +199,197 @@ extension ECDSAKey.Curve {
...
+    static var curve: ECDSAKey.Curve { get }
+}
+
+extension P256: ECDSACurveType, @unchecked @retroactive Sendable {
+    static public var curve: ECDSAKey.Curve { .p256 }
+}
+public typealias ES256PublicKey = ECDSA.PublicKey<P256>
+public typealias ES256PrivateKey = ECDSA.PrivateKey<P256>
+
+extension P384: ECDSACurveType, @unchecked @retroactive Sendable {
```
0xTim, line 213: These will need versions for Swift 5 and Swift 6

dimitribouniol: I thought it wouldn’t, because this version of the library would always be compiled with the Swift 5 language mode. Once v5 is used, then none of this is included.

dimitribouniol: Ah, nvm, forgot retroactive didn’t exist pre swift 6 compiler, will update

dimitribouniol: Though I'm now wondering if its worth adding Sendable conformances at all, since the rest of the library didn't support it yet _at all_…

0xTim: I don't mind not having them tbh, anyone wanting a full sendable supported library can bump to v5

Sources/JWTKit/JWTPayload.swift
0xTim: Nah `AsyncJWTPayload` is fine, it matches with the rest of Vapor

Sources/JWTKit/JWTPayload.swift
0xTim: I think you can mark protocol methods as deprecated can't you?

dimitribouniol: You can mark them, but you don't get any warnings telling you they are deprecated as a conformer (which makes sense since the method itself may be used by another protocol or just on its own by whoever is conforming to the protocol). I believe you do see them as a caller though, which is not as useful in this situation.

0xTim (approved):
This looks good to me, I don't know if anyone wants to comment but this is probably good to merge soon

dimitribouniol:
Still need to fix the tests, but let me get PRs for the other pieces in place before we merge so they can go in all at once

PR #228, opened 2025-05-31
Fix duplicated installation guide

fpseverino:
The installation guide in the README was duplicated.
The most recent and detailed version (which is the one left) had a typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,27 +24,27 @@ JWTKit supports all platforms supported by Swift 6 and later.
 
 ### Installation
 
-Run the following commands on your package using SwiftPM, replacing `MyTarget` with the name of your target:
-
-```sh
-cd /path/to/project/root/directory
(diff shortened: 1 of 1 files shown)


kylebrowning:
Looks like this is failing due to the expired token, I can help regenerate it, but now sure where to stick the new token, cc @0xTim

ptoffy:
@kylebrowning The JWT should replace this one https://github.com/vapor/jwt/blob/main/Tests/JWTTests/JWTTests.swift#L265 but if you prefer feel free to post it somewhere on the Vapor Discord (or DM me) and I can plug it in for you. I'd do it myself but I don't have a dev account with Sign in with Apple enabled

0xTim:
I can fix it later, it will take a day for it to expire or so

ptoffy (changes requested):
Thanks for this! I think having the "manual" way first might be better simply because I feel like it's the more common way

ptoffy (approved):

PR #229, opened 2025-06-12
Add `MLDSA{65,87}` support inside new `_QuantumJWTKit` module

ptoffy:
Since swift-crypto now supports ML-DSA{65,87}, this adds support for ML-DSA based JWTs. While ML-DSA is now a formalised standard ([RFC 204](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.204.pdf)), its usage in JOSE is still in [draft](https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/) state, which means its specifications could change. We're therefore placing it in a new, underscored (and consequently non API-stable, even though public) module called `_QuantumJWTKit`.

This PR is currently refers to the main branch of swift-crypto since no release has yet been tagged with the new post-quantum algorithms. This will change as soon as a decision is made whether to release the new APIs as a minor release _before_ the new 2026 Crypto major release, or to wait and have them in the main Crypto module. (Discussion happening [here](https://github.com/apple/swift-crypto/pull/359#issuecomment-2959350833))

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
         .library(name: "JWTKit", targets: ["JWTKit"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-crypto.git", "3.8.0"..<"5.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", branch: "main"),
         .package(url: "https://github.com/apple/swift-certificates.git", from: "1.2.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.0.0"),
     ],
@@ -27,10 +27,25 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
             ]
         ),
+        .target(
+            name: "_QuantumJWTKit",
+            dependencies: [
+                "JWTKit",
+                .product(name: "Crypto", package: "swift-crypto"),
+                .product(name: "_CryptoExtras", package: "swift-crypto"),
+            ]
+        ),
         .testTarget(
             name: "JWTKitTests",
             dependencies: [
                 "JWTKit"
+            ]
+        ),
+        .testTarget(
+            name: "_QuantumJWTKitTests",
+            dependencies: [
+                "JWTKit",
+                "_QuantumJWTKit",
             ]
         ),
     ]
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 Packages
 .build
+.index-build
 .DS_Store
 *.xcodeproj
 Package.pins
--- a/Sources/JWTKit/JWTKeyCollection.swift
+++ b/Sources/JWTKit/JWTKeyCollection.swift
@@ -32,8 +32,7 @@ public actor JWTKeyCollection: Sendable {
     public init(
         defaultJWTParser: some JWTParser = DefaultJWTParser(),
         defaultJWTSerializer: some JWTSerializer = DefaultJWTSerializer(),
-        logger: Logger = Logger(
-            label: "jwt_kit_do_not_log", factory: { _ in SwiftLogNoOpLogHandler() })
+        logger: Logger = Logger(label: "jwt_kit_do_not_log", factory: { _ in SwiftLogNoOpLogHandler() })
     ) {
         self.storage = [:]
         self.defaultJWTParser = defaultJWTParser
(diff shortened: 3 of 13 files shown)


Package.swift
```diff
@@ -27,11 +27,26 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
             ]
         ),
+        .target(
+            name: "_QuantumJWTKit",
```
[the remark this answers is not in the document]
ptoffy: Ah I didn't think about that! I had considered `_PostQuantumJWTKit` but it seems a bit long. Although yes, I think just your version is a bit less clear

fpseverino:
Why don't we hide these algorithms behind an `@_spi(PostQuantum)`?

ptoffy:
@fpseverino yes that was in fact the plan in the beginning. And nothing has changed, I just forgot 🙈 done!

0xTim:
Are we expecting these APIs to ever change? Or should we use a package trait instead of SPI?

ptoffy:
The RFC isn't final yet so even though unlikely the spec and therefore the APIs might still change

Package.swift
```diff
@@ -13,7 +13,7 @@ let package = Package(
         .library(name: "JWTKit", targets: ["JWTKit"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-crypto.git", "3.8.0"..<"5.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", branch: "main"),
```
fpseverino, line 16: [Swift Crypto v4 is now available](https://github.com/apple/swift-crypto/releases/tag/4.0.0)

ptoffy: I shall take a look this week

Sources/JWTKit/MLDSA/MLDSAError.swift
```diff
@@ -0,0 +1,5 @@
+enum MLDSAError: Error {
+    case noPrivateKey
+    case noPublicKey
```
fpseverino: This isn't used anywhere, is it?

Sources/JWTKit/MLDSA/MLDSASigner.swift
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        let signature: Data
+        do {
+            signature = try privateKey.backing.signature(for: plaintext)
+        } catch {
+            throw JWTError.signingAlgorithmFailure(MLDSAError.failedToSign(error))
+        }
+
+        return signature.copyBytes()
```
0xTim: We should investigate `Span` types for JWTKit 6

Sources/JWTKit/MLDSA/MLDSASigner.swift
```diff
@@ -0,0 +1,47 @@
...
+    init(key: Key) {
+        switch key {
+        case let key as MLDSA.PrivateKey<Key.MLDSAType>:
+            self.privateKey = key
+            self.publicKey = key.publicKey
+        case let key as MLDSA.PublicKey<Key.MLDSAType>:
+            self.privateKey = nil
+            self.publicKey = key
+        default:
+            fatalError()
```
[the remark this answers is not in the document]
ptoffy: Nah that's a cold path it'll never happen. And we'd have to make the init throwing

README.md
```diff
@@ -271,6 +271,30 @@ await keys.add(eddsa: publicKey)
 await keys.add(eddsa: privateKey)
 ```
 
+## MLDSA
+
+Hidden behind the `@_spi(PostQuantum)` flag, JWTKit supports MLDSA (Module-Lattice-Based Digital Signature Algorithm), a post-quantum signature scheme based on the CRYSTALS-DILITHIUM algorithm. It is currently behind an SPI flag because, while the MLDSA signature scheme is [standardized by NIST](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf), its [usage in JWT](https://www.ietf.org/archive/id/draft-ietf-cose-dilithium-04.html) is still in draft state, and, while unlikely, may change before being finalized. Therefore JWTKit reserves the ability to make breaking changes to this API without until the usage of MLDSA in JWT is finalized.
```
fpseverino:
```suggestion
Hidden behind the `@_spi(PostQuantum)` flag, JWTKit supports MLDSA (Module-Lattice-Based Digital Signature Algorithm), a post-quantum signature scheme based on the CRYSTALS-DILITHIUM algorithm. It is currently behind an SPI flag because, while the MLDSA signature scheme is [standardized by NIST](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf), its [usage in JWT](https://www.ietf.org/archive/id/draft-ietf-cose-dilithium-04.html) is still in draft state, and, while unlikely, may change before being finalized. Therefore JWTKit reserves the ability to make breaking changes to this API until the usage of MLDSA in JWT is finalized.
```

0xTim (approved):

PR #231, opened 2025-07-02
Fix for not honoring ValidationTimePayload's signedDate

betaphi:
# Bug Description
When using the `X5CVerifier` to verify a Payload that conforms to `ValidationTimePayload` the `signedDate` of the `ValidationTimePayload` would be used to initialize a `RFC5280Policy`. Unfortunately the default `@PolicyBuilder` that is used in the respective `verifyJWS` function also contained a `RFC5280Policy`, which was always initialized to the current date.

As both policies need to be met in order for the verification to succeed, the `ValidationTimePayload`'s `signedDate` was effectively ignored. Other negative side effects could be described but I'll keep it short here.

# Proposed Solution
As `@resultBuilder`s can not be optional in Swift, a way was needed to provide a default verification policy that would not affect the policy that would be created further down the code depending on the payload's conformance to `ValidationTimePayload`. I have therefore created a dummy `EmptyPolicy`, which is always met.
The solution does not have any negative side effects, as a `RFC5280Policy` is added further down the code in any case. Only this time, it is either initialized with the `signedDate` of a `ValidationTimePayload` or the current date.

### DIFF
--- a/Sources/JWTKit/X5C/EmptyPolicy.swift
+++ b/Sources/JWTKit/X5C/EmptyPolicy.swift
@@ -1,0 +1,24 @@
+//
+//  AlwaysMeetsPolicy.swift
+//  jwt-kit
+//
+//  Created by Bastian Rössler on 02.07.25.
+//
+
+import X509
+import SwiftASN1
+
+/// This Policy acts as a placeholder. Its result is always positive.
+@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, macCatalyst 13, visionOS 1.0, *)
+public struct EmptyPolicy: VerifierPolicy {
+    @inlinable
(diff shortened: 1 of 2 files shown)


0xTim:
Thanks! Can you add a test to ensure that the original bug doesn't get reintroduced?

ptoffy (changes requested):
Thanks! Unfortunately we're not quite there yet. Could you fix the compiler error please? Also, this needs to go through a `swift format` pass first (formatting file is [here](https://github.com/vapor/contributing/blob/main/.swift-format))

betaphi:
The compiler error was caused by an update to a dependency library. I just fixed it.

About the `swift format` thing: The whole repo does not comply to the current `format`. As a result, if I apply it, lots of files will be changed. Therefore I would highly suggest to do this in a separate PR.

PR #232, opened 2025-10-09
Finally remove `@retroactive @unchecked Sendable` from crypto types

ptoffy:


ptoffy:
Merging over the API checker which isn't recognising that there's a typealias between the old and new return types.

PR #235, opened 2025-10-22
Add Support for SecureEnclave.P256

jseibert:
This PR adds support for keys managed by the SecureEnclave on iOS/macOS:

```swift
let privateKey = SecureEnclave.P256.Signing.PrivateKey()
let keys = JWTKeyCollection()
await keys.add(ecdsa: ES256EnclavePrivateKey(backing: privateKey))
let jwt = try await keys.sign(payload)
print("JWT:", jwt)
```

I would love feedback on approach. This was the minimal set of changes I was able to make to add support to the existing library structure.

### DIFF
--- a/Sources/JWTKit/ECDSA/ECDSA.swift
+++ b/Sources/JWTKit/ECDSA/ECDSA.swift
@@ -16,6 +16,14 @@ public protocol ECDSAKey: Sendable {
     var parameters: ECDSAParameters? { get }
 }
 
+public protocol ECDSAEnclaveKey: Sendable {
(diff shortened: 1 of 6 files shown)


0xTim:
@jseibert is the aim to add support for Apple platforms?

Assuming that's the case, there's a few things here. Number one, any mention on the Secure Enclave needs to be gated behind availability checks since it's not available on non-Apple platforms. Second, we can't break the API by changing protocols etc. It should be it's own standalone type

Ari4ka (approved):

PR #237, opened 2025-10-27
Add PEM representation init and property to EdDSA keys

ptoffy:
This adds PEM representation support for EdDSA keys, it's now possible to create EdDSA keys with PEM and get their PEM representation. 
This also adds the missing `Equatable` conformance to EdDSA keys and fixes some tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,7 +32,7 @@ jobs:
       with_api_check: ${{ github.event_name == 'pull_request' }}
       warnings_as_errors: true
       with_linting: true
-      with_windows: true
+      with_windows: false
       with_musl: true
       with_android: true
       ios_scheme_name: jwt-kit


Sources/JWTKit/EdDSA/EdDSA.swift
```diff
@@ -19,7 +19,7 @@ extension EdDSA {
     ///
     /// In JWT, EdDSA public keys are represented as a single x-coordinate and are used for verifying signatures.
     /// Currently, only the ``EdDSACurve/ed25519`` curve is supported.
-    public struct PublicKey: EdDSAKey {
+    public struct PublicKey: EdDSAKey, Equatable {
```
tkrajacic: Out of curiosity, why can't we also make them Hashable?

ptoffy: We could if there's a use case for it. I wouldn't have made them `Equatable` either if 
1. We didn't need it in the tests
2. RSA and ECDSA weren't also already `Equatable`

tkrajacic: I think all keys should conform to `Hashable` because… they are

PR #243, opened 2026-04-28
Remove NSString API usage

### LINKED ISSUE #242: Remove NSString API usage
Foundation is still used through these functions: https://github.com/vapor/jwt-kit/blob/aa60a211797306bfb05b752ad7b4bf9f0b50d898/Sources/JWTKit/Utilities/Base64URL.swift#L9-L10

Since it's an extension for a stdlib type, it's enough that Foundation is imported in a dependency but if all of the dependencies remove all Foundation imports, it will stop compiling.

Cyberbeni:
**These changes are now available in [5.4.1](https://github.com/vapor/jwt-kit/releases/tag/5.4.1)**

resolves #242 

Remove NSString API usage

### DIFF
--- a/Sources/JWTKit/Utilities/Base64URL.swift
+++ b/Sources/JWTKit/Utilities/Base64URL.swift
@@ -6,8 +6,8 @@ import Foundation
 
 extension String {
     package func base64URLDecodedData() -> Data? {
-        var base64URL = replacingOccurrences(of: "-", with: "+")
-            .replacingOccurrences(of: "_", with: "/")
+        var base64URL = replacing("-", with: "+")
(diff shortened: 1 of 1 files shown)


ptoffy (approved):

ptoffy (changes requested):
Actually this won't work in iOS 15. We have a native base64URL unescaping function already, let's use it
```swift
extension String {
    package func base64URLDecodedData() -> Data? {
        var data = Data(self.utf8)
        data.base64URLUnescape()
        return Data(base64Encoded: data)
    }
}
```

ptoffy (approved):

PR #244, opened 2026-05-07
X5CVerifier: support for RSA

m-barthelemy:
**These changes are now available in [5.6.0](https://github.com/vapor/jwt-kit/releases/tag/5.6.0)**

After https://github.com/vapor/jwt-kit/pull/238 adding support for .es256, .es384, .es512 and  .eddsa, this PR adds support for RSA signatures.

### DIFF
--- a/Sources/JWTKit/X5C/X5CVerifier.swift
+++ b/Sources/JWTKit/X5C/X5CVerifier.swift
@@ -221,6 +221,39 @@ extension X5CVerifier {
         case .eddsa:
             let eddsaKey = try EdDSA.PublicKey(pem: certificate.publicKey.serializeAsPEM().pemString)
             return JWTSigner(algorithm: EdDSASigner(key: eddsaKey), parser: parser)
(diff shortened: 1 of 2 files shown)


m-barthelemy:
Hi @ptoffy would you like this to be handled differently?

ptoffy:
Hey @m-barthelemy, thanks for the ping and sorry, busy times. I'm usually quite wary about adding anything RSA related, but since we provide RSA keys already I think this addition does make sense. Thanks!

ptoffy (approved):

m-barthelemy:
> Hey @m-barthelemy, thanks for the ping and sorry, busy times. 

No worries, busy times here too. Thanks a lot for merging this!

> I'm usually quite wary about adding anything RSA related, but since we provide RSA keys already I think this addition does make sense.

I would not use RSA on a new project either, in fact what made me create this PR is the [Fido Metadata](https://fidoalliance.org/metadata/) - they issue blobs signed using an RSA cert.

PR #247, opened 2026-07-28
Use comparison operator instead of compare function

madrob:
Use modern Swift less-than operator instead of Obj-C style compare function for date check.


ptoffy (approved):
Thanks!

PR #250, opened 2026-08-05
Assert HMAC key size

ptoffy:
According to [RFC 7518 section 3.2](https://datatracker.ietf.org/doc/html/rfc7518#section-3.2):
> A key of the same size as the hash output (for instance, 256 bits for "HS256") or larger MUST be used with this algorithm.

Since we cannot make the initialiser throwing, and since we might break some deployments if we used `precondition`, `assert` at least nudges people towards the correct key size, until we can enforce this with an error in a future major.

### DIFF
--- a/Sources/JWTKit/HMAC/HMACSigner.swift
+++ b/Sources/JWTKit/HMAC/HMACSigner.swift
@@ -13,7 +13,7 @@ struct HMACSigner<SHAType>: JWTAlgorithm where SHAType: HashFunction {
     init(key: SymmetricKey) {
         assert(
             key.bitCount >= SHAType.Digest.byteCount * 8,
-            "Key should be at least as large as the hash output: \(SHAType.Digest.byteCount) bytes"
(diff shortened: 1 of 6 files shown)


0xTim (approved):
To be clear we _can_ switch to a precondition without a SemVer major, breaking behaviour changes are allowed but we should give people time to migrate over at the very least.

Sources/JWTKit/HMAC/HMACSigner.swift
```diff
@@ -11,6 +11,10 @@ struct HMACSigner<SHAType>: JWTAlgorithm where SHAType: HashFunction {
     let name: String
 
     init(key: SymmetricKey) {
+        assert(
+            key.bitCount >= SHAType.Digest.byteCount * 8,
+            "Key should be at least as large as the hash output: \(SHAType.Digest.byteCount) bytes"
```
0xTim: Add that this will become a precondition in a future release

0xTim (changes requested):
One nit

0xTim (approved):

PR #252, opened 2026-08-17
Make MLDSA APIs part of the public API

ptoffy:
The ML-DSA JOSE proposal is not in draft state anymore (https://datatracker.ietf.org/doc/rfc9964/), so let's welcome it to our public API


fpseverino (approved):
LGTM

PR #256, opened 2026-09-09
Add instructions count to benchmarks

ptoffy:


Benchmarks/Signing/Signing.swift
```diff
@@ -5,8 +5,12 @@ import Utilities
 
 let benchmarks = {
     Benchmark.defaultConfiguration = .init(
-        metrics: [.peakMemoryResident, .mallocCountTotal],
+        metrics: [.instructions, .mallocCountTotal, .peakMemoryResident],
```
MahdiBM, line 14: It doesn't really matter for our case of catching regressions or improvements, but to be clear, the instruction-count benchmarks will show higher numbers than the actual code does because counting mallocs requires replacing the allocator funcs with some custom code.

MahdiBM (approved):
Should there be some threshold file updates?

Benchmarks/Signing/Signing.swift
ptoffy: Good to know!

PR #258, opened 2026-09-10
Improve verification performance

ptoffy:
We were doing a lot of useless parsing: 
- `JWTHeaderField` doesn't need to be `indirect`
- `JWTHeaderField` parsing is very slow due to trying various types, so we re-order type decoding to most likely (String) first
- Avoid parsing the whole header when we only need specific parts and decode those directly
- Only split the token once at the beginning and thread down the single parts


MahdiBM:
@penny benchmark

ptoffy:
@MahdiBM wrong penny 😂

MahdiBM:
I know 😆. It'll still work though when the PR is merged.

That comment was just to gather data for tests.

ptoffy:
@penny-for-vapor benchmark

MahdiBM:
@Penny benchmark

MahdiBM (approved):

ptoffy:
@penny-for-vapor benchmark

ptoffy:
@penny-for-vapor benchmark

PR #260, opened 2026-09-11
Add 6.4 specific manifest

### LINKED ISSUE #259: Vapor project build fails after updating JWTKit to 5.7.0
### Describe the issue

The latest release of jwt-kit (5.7.0) adds warnings-as-errors which, due to a SwiftPM bug, causes building to fail if your project doesn't also use that flag.

### Vapor version

4.122.0

macOS 26.6.2

### Swift version

Swift Package Manager - Swift 6.3.0-dev

### Steps to reproduce

1. Package has Vapor as dependency, which imports JWTKit
2. Update JWTKit to v5.7.0, which adds the warnings-as-errors flag
3. Build package in Xcode

### Outcome

Build fails, likely due to [this SwiftPM bug](https://github.com/swiftlang/swift-package-manager/issues/9517).
Pinning JWTKit to 5.6.0 resolves this issue.

### Additional notes

_No response_

ptoffy:
**These changes are now available in [5.7.1](https://github.com/vapor/jwt-kit/releases/tag/5.7.1)**

Fixes #259

### DIFF
--- a/Package@swift-6.4.swift
+++ b/Package@swift-6.4.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:6.2
+// swift-tools-version:6.4
 import PackageDescription
 
 let package = Package(
--- a/Package@swift-6.4.swift
+++ b/Package@swift-6.4.swift
@@ -40,6 +40,8 @@ let package = Package(
 
 var swiftSettings: [SwiftSetting] {
     [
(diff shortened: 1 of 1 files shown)


MahdiBM (commented):
This is ok but I think when e.g. Swift 6.5 comes out it won't use the 6.4 package manifest? I'm not sure. If that's true, we _could_ instead have a 6.3 manifest and symlink it to a 6.2 file and let the main manifest be for 6.4 and above.

